PR #1, opened 2020-06-01
feat: Initial proof of concept

msbrogli:
Requires this MR: https://gitlab.com/HathorNetwork/hathor-python/-/merge_requests/406

Design: https://github.com/HathorNetwork/rfcs/issues/24

To do list:
- [x] OpenAPI Docs


txstratum/client.py
```diff
@@ -0,0 +1,111 @@
...
+            txout = TxOutput(
+                value=do['value'],
+                token_data=0,
+                script=b'',
+            )
+            blk.outputs = [txout]
+            return BlockTemplate(data=bytes(blk), height=height)
+
+    async def get_tx_parents(self) -> List[bytes]:
+        """Return parents for a new transactions."""
```
pedroferreira1, line 98: for a new **transaction**

txstratum/client.py
```diff
@@ -0,0 +1,111 @@
...
+            major, minor, patch = ver.split('.')
+            return HathorVersion(int(major), int(minor), int(patch))
+
+    async def get_block_template(self) -> BlockTemplate:
+        """Return a block template."""
+        assert self._session is not None
+        async with self._session.get(self._get_url('get_block_template')) as resp:
+            data = await resp.json()
+            # TODO Get block's height.
+            height = 0
```
pedroferreira1, line 79: Is this height that will come from the full node?

txstratum/api.py
```diff
@@ -0,0 +1,112 @@
...
+        Format: json
+        Params:
+        - tx: str, hex dump of the transaction
+        - propagate: bool, propagate tx to Hathor’s full node after it is solved
+        - add_parents: bool, add parents before resolving the tx
+        """
+        try:
+            data = await request.json()
+        except json.decoder.JSONDecodeError:
+            return web.json_response({'error': 'cannot decode json'}, status=400)
```
pedroferreira1, line 55: Should we have error codes? Like the ones we've discussed to add in the full node

txstratum/jobs.py
```diff
@@ -0,0 +1,171 @@
...
+            expected_total_time: int, sum of expected_queue_time and expected_mining_time (in seconds)
+        """
+        return {
+            'job_id': self.uuid.hex(),
+            'status': self.status.value,
+            'created_at': self.created_at,
+            'nonce': self.nonce.hex() if self.nonce else None,
+            'submitted_at': self.submitted_at,
+            'total_time': self.total_time,
+            # TODO
```
pedroferreira1, line 123: Important TODO because the estimation values are still hardcoded

txstratum/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Copyright (c) Hathor Labs and its affiliates.
```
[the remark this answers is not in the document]
msbrogli: We are. I'll double check and fix them. Thanks!

txstratum/api.py
```diff
@@ -0,0 +1,112 @@
...
+        try:
+            tx_bytes = bytes.fromhex(tx_hex)
+        except ValueError:
+            return web.json_response({'error': 'invalid tx format'}, status=400)
+        try:
+            Transaction.create_from_struct(tx_bytes)
+        except ValueError:
+            return web.json_response({'error': 'invalid tx'}, status=400)
+        add_parents = data.get('add_parents', False)
+        # propagate: bool
```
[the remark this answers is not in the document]
msbrogli: You're right. I'll fix it. Thanks.

txstratum/api.py
```diff
@@ -0,0 +1,112 @@
...
+
+    async def cancel_job(self, request: web.Request) -> web.Response:
+        """Cancel a tx job.
+
+        Method: POST
+        Query:
+        - job-id: str, job identifier
+        """
+        uuid_hex = request.query.get('job-id')
+        if not uuid_hex:
```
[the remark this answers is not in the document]
msbrogli: I'll do that. Thanks!

txstratum/client.py
msbrogli: It is. I know it is somewhere in the `data` object.

txstratum/api.py
msbrogli: It is a good idea. I'll add it. Thanks!

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  - pip install pipenv
+  - pipenv install --ignore-pipfile --deploy --system --dev
+
+jobs:
+  allow_failures:
+    - stage: check
+      script: make check
+  include:
+    - stage: tests
+      script: make tests
```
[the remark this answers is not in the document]
msbrogli: I tried to put them in different stages. But it is not working. I'll check the docs to find what is wrong.

txstratum/__init__.py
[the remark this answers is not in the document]
msbrogli: It was missing in the test files. Thanks!

txstratum/api.py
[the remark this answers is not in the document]
msbrogli: Refactored. Thanks!

txstratum/client.py
msbrogli: Fixed. Thanks!

.travis.yml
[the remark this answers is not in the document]
msbrogli: Fixed.

txstratum/protocol.py
```diff
@@ -0,0 +1,345 @@
...
+
+        now = time.time()
+        self.completed_jobs += 1
+        job.submitted_at = now
+        self.last_submit_at = now
+        self._submitted_work.append(SubmittedWork(job.submitted_at, job.share_weight))
+        self.send_result(msgid, 'ok')
+
+        # XXX Is it a good rule when mining txs?
+        # Too many jobs too fast, increase difficulty out of caution (more than 10 submits within the last 10s)
```
obiyankenobi, line 212: I'm not sure if we should do that.

For blocks, it shouldn't happen. We probably won't have powerful miners connected to this API that could solve blocks very fast and, even if we did, weight adjustment should take care of it.

For transactions, I think it does not make sense.

txstratum/manager.py
```diff
@@ -0,0 +1,206 @@
...
+                self.update_miner_job(protocol)
+
+    def _job_clean_up(self, job: MinerTxJob) -> None:
+        """Clean up tx job. It is scheduled after a tx is solved."""
+        self.tx_jobs.pop(job.uuid)
+
+    def get_best_job(self, protocol: StratumProtocol) -> MinerJob:
+        """Return best job for a miner."""
+        if len(self.tx_queue) > 0:
+            return self.tx_queue.popleft()
```
obiyankenobi, line 204: I'm not sure this is the exact point where we would implement these changes, but I thought it was best to comment here so we could keep all in a thread (as opposed to a single comment on the PR).

There are 2 things I've seen are important for mining transactions and I think haven't been implemented:
- adding a timeout. In our latest implementation we use 1s;
- update tx timestamp/parents when sending the job to the miner. This ensures miners get new data to mine on;

What we'd see is that sometimes the miners would not find the correct nonce for a transaction, even with low weight (let's say 20) in a reasonable amount of time. That's why it's important to have the timeout and to update the tx, otherwise the miner would mine over the same piece of data.

txstratum/api.py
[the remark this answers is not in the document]
msbrogli: Done!

txstratum/api.py
msbrogli: I used the notation cited by @obiyankenobi. Would you check again, please?

README.md
```diff
@@ -0,0 +1,17 @@
+# tx-mining-service
+
+The tx-mining-service is a service used to resolving transactions before propagating them into Hathor's mainnet.
+
+It has two kind of clients: users and miners. Users use it when sending new transactions; while miners do the work of resolving the transactions.
```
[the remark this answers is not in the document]
msbrogli: Fixed. Thanks!

tests/commons.py
```diff
@@ -0,0 +1,56 @@
...
+                             '840a6dd5d4e4e600d2da8d')
+        tx = Transaction.create_from_struct(data)
+
+        self.assertEqual(data, bytes(tx))
+        self.assertTrue(tx.verify_pow())
+        self.assertTrue(tx.is_transaction)
+        self.assertFalse(tx.is_block)
+
+        print(str(tx))
+        print(repr(tx))
```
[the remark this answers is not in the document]
msbrogli: They are there to test the methods `str()` and `repr()`.

txstratum/api.py
```diff
@@ -0,0 +1,108 @@
...
+            return web.json_response({'error': 'cannot decode json'}, status=400)
+        if not isinstance(data, dict):
+            return web.json_response({'error': 'json must be an object'}, status=400)
+        tx_hex = data.get('tx')
+        if not tx_hex:
+            return web.json_response({'error': 'missing tx'}, status=400)
+        try:
+            tx_bytes = bytes.fromhex(tx_hex)
+        except ValueError:
+            return web.json_response({'error': 'invalid tx format'}, status=400)
```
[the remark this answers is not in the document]
msbrogli: Done. Thanks!

txstratum/api.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import TYPE_CHECKING, Optional
+
+from aiohttp import web
+from txstratum.commons import Transaction
+from txstratum.jobs import MinerTxJob
+
+if TYPE_CHECKING:
+    from txstratum.manager import TxMiningManager
+
+routes = web.RouteTableDef()
```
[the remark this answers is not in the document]
msbrogli: Fixed. Thanks!

txstratum/manager.py
msbrogli: I've implemented the update timestamp for both tx and blocks. What do you mean by a timeout? We already update the job every 30 seconds, updating both the timestamp and the weight. The target time for blocks is 15 seconds.

Should I implement different job update interval? Maybe 30 seconds for blocks and 5 seconds for transactions?

tests/commons.py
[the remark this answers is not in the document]
obiyankenobi: Ah, ok

txstratum/manager.py
obiyankenobi: We've been using 1 second timeouts for tx jobs. We had a discussion here: https://gitlab.com/HathorNetwork/hathor-python/-/merge_requests/398#note_339321295

tests/commons.py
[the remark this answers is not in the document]
jansegre: But then it just tests it's not raising an exception, right? isn't it better to use `self.assertEqual` with the expected output of `str()` and `repr()`?

txstratum/client.py
```diff
@@ -0,0 +1,112 @@
+"""
```
jansegre, line 1: Shouldn't this be in `commons` too? We could later turn commons into a lib that merged mining can use too.

txstratum/client.py
jansegre: It should be on `data.metadata.height`.

msbrogli: Done. Thanks!

txstratum/jobs.py
msbrogli: Done. Thanks!

txstratum/client.py
msbrogli: I can move it there. Maybe the `commons/` folder should become a `hathor-commons` package itself.

txstratum/protocol.py
```diff
@@ -0,0 +1,351 @@
...
+
+        self._submitted_work: List[SubmittedWork] = []
+
+        self._authorized: bool = False
+        self._subscribed: bool = False
+        self.started_at: float = 0.0
+        self.estimator_task: Optional[Periodic] = None
+
+        # For testing only.
+        self._update_job_timestamp: bool = True
```
obiyankenobi, line 77: Why is this just for testing?

txstratum/api.py
```diff
@@ -0,0 +1,109 @@
...
+        if not tx_hex:
+            return web.json_response({'error': 'missing-tx'}, status=400)
+        try:
+            tx_bytes = bytes.fromhex(tx_hex)
+        except ValueError:
+            return web.json_response({'error': 'invalid-tx'}, status=400)
+        try:
+            Transaction.create_from_struct(tx_bytes)
+        except ValueError:
+            return web.json_response({'error': 'invalid-tx'}, status=400)
```
obiyankenobi, line 67: Can we merge this?
```
try:
    tx_bytes = bytes.fromhex(tx_hex)
    Transaction.create_from_struct(tx_bytes)
except ValueError:
    return web.json_response({'error': 'invalid-tx'}, status=400)
```

txstratum/api.py
```diff
@@ -0,0 +1,109 @@
...
+
+        add_parents = data.get('add_parents', False)
+        propagate = data.get('propagate', False)
+        job = MinerTxJob(tx_bytes, add_parents=add_parents, propagate=propagate)
+        status = self.manager.add_job(job)
+        if status is None:
```
obiyankenobi, line 73: `add_job` returns a boolean.

txstratum/api.py
obiyankenobi: Good for me.

txstratum/cli.py
```diff
@@ -0,0 +1,81 @@
...
+    parser.add_argument('backend', help='Endpoint of the Hathor API (without version)', type=str)
+    return parser
+
+
+def execute(args: Namespace) -> None:
+    """Run the service according to the args."""
+    from txstratum.client import HathorClient
+
+    logger.info('Running tx-mining-service...')
+    logger.info('Backend: {}'.format(args.backend))
```
jansegre, line 38: These are called before `logging.config` so they won't show up.

Pipfile
```diff
@@ -0,0 +1,24 @@
+[[source]]
```
jansegre, line 1: *pretend I'm commenting the whole file*

I think `Pipfile.lock` should be commited because this is an application not a library, so the exact versions matter a lot.

Also, consider [Poetry](https://python-poetry.org/), I have [a branch on hathor-core](https://github.com/HathorNetwork/hathor-core/tree/chore/replace-pipenv-with-poetry) with it, and it's very similar to setup. If we want to move away from Pipenv, it's a good thing to test the alternatives on smaller deploys.

tests/commons.py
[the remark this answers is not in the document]
msbrogli: I agree. I just changed to `self.assertEqual`. Would you review again, please?

txstratum/protocol.py
msbrogli: Do you think we should use only blocks to estimate the miner's hash rate?

txstratum/manager.py
msbrogli: I implemented a 2 second update. What do you think?

txstratum/protocol.py
msbrogli: I disable updating job timestamps to make testing easier. The issue was comparing two jobs there are supposed to be the same but they have different timestamps. But maybe I can use the hash of the funds when comparing. What do you think?

txstratum/api.py
msbrogli: We can. Just did it. Thanks!

txstratum/client.py
msbrogli: I haven't moved it yet because it is using structlog. How can we generalize the logging? Just use `import logging` instead?

txstratum/cli.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def create_parser() -> ArgumentParser:
+    """Create a parser for the cmdline arguments."""
+    import configargparse  # type: ignore
+    parser: ArgumentParser = configargparse.ArgumentParser(auto_env_var_prefix='hathor_')
+    parser.add_argument('--stratum-port', help='Port of Stratum server', type=int, default=8000)
+    parser.add_argument('--api-port', help='Port of TxMining API server', type=int, default=8080)
+    parser.add_argument('--log-config', help='Config file for logging', default='log.conf')
+    parser.add_argument('--max-tx-weight', help='Maximum allowed tx weight to be mined', type=int, default=None)
```
msbrogli, line 28: I added both `--max-tx-weight` and `--tx-timeout`. @jansegre @obiyankenobi @pedroferreira1 Would you please review it? I put in a separate commit to make it easier to see the changes.

txstratum/protocol.py
obiyankenobi: Yes, as we've had the random mining time for transactions, we should only use blocks for hash rate estimation.

txstratum/protocol.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def start_estimator(self) -> None:
+        """Start periodic estimator task."""
+        if self.estimator_task is None:
+            self.last_reduced = time.time()
+            # Start estimator periodic task.
+            self.estimator_task = Periodic(self.estimator_loop, self.ESTIMATOR_LOOP_INTERVAL)
+            asyncio.ensure_future(self.estimator_task.start())
+            # Start fresh job periodic task.
+            self.refresh_job_task = Periodic(self.refresh_job, self.JOB_UPDATE_INTERVAL)
```
obiyankenobi, line 135: This will run every 2 seconds, right? There are 2 implications:
1. For blocks, we don't need to update the job every 2 seconds. I'm not sure if it impacts performance (and in the end hash rate estimation), but as we're not concerned with mining blocks here, it's probably not a problem;
2. Ideally, we should start the 2s timer when we send the new tx job to the miner. Since this runs every 2 seconds regardless of when the job starts, it might happen that it'll update the job after it's been running for a few milliseconds. After this first one, the job send interval should "sync" with this periodic task, so maybe there's not much impact.

txstratum/client.py
jansegre: I think it's fine if it depends on structlog. At the very least we can think of replacing it when moving it out from the project.

txstratum/protocol.py
msbrogli: Sorry, my explanation wasn't clear. I meant that I disabled updating job timestamp for some of the unittests.

txstratum/api.py
msbrogli: Fixed. Thanks!

txstratum/cli.py
msbrogli: Fixed. Thanks!

Pipfile
msbrogli: I'll do it. Thanks!

txstratum/protocol.py
```diff
@@ -0,0 +1,358 @@
...
+            asyncio.ensure_future(self.refresh_job_task.stop())
+
+    def start_estimator(self) -> None:
+        """Start periodic estimator task."""
+        if self.estimator_task is None:
+            self.last_reduced = time.time()
+            # Start estimator periodic task.
+            self.estimator_task = Periodic(self.estimator_loop, self.ESTIMATOR_LOOP_INTERVAL)
+            asyncio.ensure_future(self.estimator_task.start())
+            # Start fresh job periodic task.
```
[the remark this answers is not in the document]
msbrogli: Renamed to `start_periodic_tasks`. Thanks!

txstratum/protocol.py
msbrogli: 1) It shouldn't impact performance because we don't send `clean=True`. So, the client can decide the best cycle to update the job.

2) As we are using `clean=False`, it seems ok. It might happen that a miner will receive a job and an update just after, but it shouldn't affect the overall performance.

Pipfile
msbrogli: I kept Pipfile for now. But I'll open an issue to migrate to Poetry soon. Ok?

txstratum/protocol.py
obiyankenobi: Ah, got it.

txstratum/manager.py
obiyankenobi: 2 seconds seems good for now. We might want to have an adjustable (per miner) interval in the future. When I did some testing in our stratum, my miner was better with 1s intervals (the internet here is pretty reliable). But I understand Jan had a bigger latency, so 1s might be too quick for him.

Pipfile
jansegre: OK, no problem.

txstratum/cli.py
jansegre: I don't think it's super important for `max-tx-weight` to be configurable, but it's fine. But I think it should have a reasonable default. It certainly doesn't make sense to allow impossibly large tx.weights. I'd be OK if we just estimated an upper bound for tx.weight and made that hardcoded.

`--tx-timeout` looks great though.

txstratum/protocol.py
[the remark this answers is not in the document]
obiyankenobi: This is still under `if self.estimator_task is None:`. Should it be under `if self. refresh_job_task is None:`?

txstratum/api.py
```diff
@@ -17,13 +15,24 @@
     from txstratum.manager import TxMiningManager
 
 
+# Default maximum tx weight allowed to be mined in this server.
+MAX_TX_WEIGHT = 35
```
pedroferreira1: We should create a task on hathor-core to update our method that calculates the tx weight, so it can't be above 35.

txstratum/manager.py
```diff
@@ -52,6 +50,11 @@ def __init__(self, backend: 'HathorClient'):
         self.block_template_updated_at: float = 0
         self.block_template: Optional['BlockTemplate'] = None
 
+        # Statistics
+        self.txs_solved: int = 0
+        self.blocks_found: int = 0
+        self.txs_timeout: int = 0
```
pedroferreira1, line 56: Not right now but I think we could create a task to add this info to prometheus, so we can add this to grafana if we want. 

But maybe we could have this on a `status` API

txstratum/cli.py
jansegre: I just noticed that `--max-tx-weight` defaults to `MAX_TX_WEIGHT` when it's `None`. I think it's fine then.

msbrogli: Ok!

txstratum/client.py
msbrogli: Ok. Done!

txstratum/protocol.py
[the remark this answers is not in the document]
msbrogli: You're right. Fixed. Thanks!

txstratum/api.py
msbrogli: I think this is just to prevent abusing our service, so I don't think we need to hard cap the tx weight in the full node.

txstratum/manager.py
msbrogli: I added them to the `status` API. I will also add these metrics to our Grafana in the next PR. I'll focus on deployment and monitoring.

txstratum/protocol.py
obiyankenobi: @msbrogli I still think this should be removed.

msbrogli: Done!

msbrogli: The changes are in the last commit. Can you review it, please? @obiyankenobi @pedroferreira1 @jansegre

obiyankenobi: Looks good to me.

txstratum/api.py
pedroferreira1: Yeah, I agree but there are some legit transactions that won't be able to be mined by this service with this cap, that's the problem I see. But I think it's okay we should just know that this case might happen and the user must resolve this tx elsewhere and propagate to the full node later

msbrogli: We haven't seen a transaction with such weight so far. But it is a minimum tx weight, so maybe we can even reduce that number in the future.

pedroferreira1: Yeah, I think it's fine.

PR #3, opened 2020-08-04
feat(monitor): Add Prometheus metrics exporter

msbrogli:


txstratum/prometheus.py
```diff
@@ -0,0 +1,90 @@
...
+from typing import TYPE_CHECKING, Dict
+
+from prometheus_client import CollectorRegistry, Gauge, write_to_textfile  # type: ignore
+
+if TYPE_CHECKING:
+    from txstratum.manager import TxMiningManager
+
+
+# Define prometheus metrics and it's explanation
+METRIC_INFO = {
```
obiyankenobi, line 12: It's probably good to add a comment here that these need to be a subset of `manager.status()`.

It might also be good to have a comment on `manager.status()` explaining this, otherwise someone might change some metrics names and it'll break things here.

txstratum/manager.py
```diff
@@ -136,6 +137,7 @@ async def update_block_template(self) -> None:
         except Exception:
             # XXX What should we do?!
             self.block_template = None
+            self.block_template_error += 1
```
pedroferreira1, line 140: I know it's not from this PR but I was gonna ask to log something here because we are capturing all exceptions, so we could have the log of it. But given that we still don't have logging here, maybe just create an issue to do this in the future?

txstratum/prometheus.py
```diff
@@ -0,0 +1,90 @@
...
+}
+
+
+class PrometheusExporter:
+    """Class that sends hathor metrics to a node exporter that will be read by Prometheus."""
+
+    def __init__(self, manager: 'TxMiningManager', path: str, filename: str = 'tx-mining-service.prom'):
+        """Init PrometheusExporter.
+
+        :param metrics: Metric object that stores all the hathor metrics
```
pedroferreira1, line 30: Wrong comment, this should be manager object. Maybe we could remove this, since we already have the type on the params.

pedroferreira1, line 30: Wrong comment, this should be manager object. Maybe we could remove this, since we already have the type on the params.

msbrogli: Fixed. Thanks!

txstratum/manager.py
msbrogli: Just added it. Thanks!

txstratum/prometheus.py
msbrogli: I changed it and added a simple test. Can you review it again, please?

obiyankenobi: To be honest, I think it became more complex now, with the lambdas on `METRIC_INFO`. I think the old way was easier to understand. Maybe I'm not seeing something.

I did like the idea of having a test for it, instead of relying on the comments, as I suggested. If we decide to go back to the old way, we could just check that `METRIC_INFO.keys()` is a subset of `manager.status().keys()`.

msbrogli: I changed it because the `manager.status()` was calculating more things than needed, e.g. generating the status of the miners. This way, we can explicitly add whatever metric we want regardless of the public status.

obiyankenobi: Ah, makes sense!

msbrogli: Another possibility would be to create a `get_metrics()` method in the manager. So, the PrometheusExporter would just call it. What do you think?

obiyankenobi: That might be a bit cleaner than using the lambdas on `METRIC_INFO`.

msbrogli: Done!

PR #5, opened 2020-12-17
feat: create conf files for testnet and mainnet, so it can easily select the one to connect

pedroferreira1:


obiyankenobi (approved):

jansegre (approved):
I don't really like the mechanism we use for having multiple configs, but it's fine for now.

PR #7, opened 2021-01-13
docs: Fix apidocs body parameters to work with openapi 3.0

andreabadesso:
Fixed the apidocs schema and added json and yaml files to make it easier to import on http clients


msbrogli:
@andreabadesso Can you rebase, please? It should pass all tests.

api-docs.js
```diff
@@ -103,43 +103,38 @@ const apiDoc = {
     '/submit-job': {
```
msbrogli, line 103: Can you add the script used to generate the json and yaml files? I agree that it is easier to have those docs committed but I'm wondering how we protect from updating the `api-docs.js` and forgetting to update the generated files. Any suggestion?

msbrogli (changes requested):

jansegre (approved):
I think `docs/apidocs.yml` should be enough. Is there any benefit in including `docs/apidocs.json`? Either way this looks good.

pedroferreira1 (approved):

api-docs.js
andreabadesso: I generated them manually using the `swagger editor`.

From what I understand, the only reason we have a `.js` file in this project is to be able to add comments to the docs

What I suggest is removing the api-docs.js file and keeping only `docs/apidocs.yml` as yaml supports comments.

In other javascript projects, we could parse the yaml file to use it on javascript (for a http route or something like this).

What do you think?

andreabadesso:
> @andreabadesso Can you rebase, please? It should pass all tests.

Done

api-docs.js
msbrogli: Ok. I agree that keeping only `docs/apidocs.yml` is better and more handy.

andreabadesso: @msbrogli Done!

msbrogli (approved):

PR #8, opened 2021-01-15
fix: Only consider positive mining time when estimating queue time

msbrogli:
When a transaction is received to be mined, the tx mining service estimates how long it will take to get it done and saves it to `job.expected_mining_time`. If this time cannot be calculated, this value is set to `-1`.

Finally, when there's a queue of transactions, we sum the individual `expected_mining_time` of the transactions to calculate the `job.expected_queue_time`. In this case, we must ignore the transactions with `-1` value.

We found out this bug when many transactions had `expected_mining_time = -1` and new transactions were summing and setting the `expected_queue_time` to negative values (e.g., -7, -9, -17).

### DIFF
--- a/txstratum/manager.py
+++ b/txstratum/manager.py
@@ -237,7 +237,9 @@ class TxMiningManager:
         """Enqueue a tx job to be mined."""
         assert job not in self.tx_queue
         job.status = JobStatus.ENQUEUED
-        job.expected_queue_time = sum(x.expected_mining_time for x in self.tx_queue)
+        # When the total hashrate is unknown, `expected_mining_time` is set to -1. Thus, we need to
+        # skip negative values when calculating the `expected_queue_time`.
+        job.expected_queue_time = sum(x.expected_mining_time for x in self.tx_queue if x.expected_mining_time > 0)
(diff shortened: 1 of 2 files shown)


jansegre (approved):

pedroferreira1 (approved):

txstratum/manager.py
```diff
@@ -237,7 +237,7 @@ def enqueue_tx_job(self, job: MinerTxJob) -> None:
         """Enqueue a tx job to be mined."""
         assert job not in self.tx_queue
         job.status = JobStatus.ENQUEUED
-        job.expected_queue_time = sum(x.expected_mining_time for x in self.tx_queue)
+        job.expected_queue_time = sum(x.expected_mining_time for x in self.tx_queue if x.expected_mining_time > 0)
```
pedroferreira1, line 240: Maybe we could add the same description you used in the PR here as a comment. What do you think?

pedroferreira1 (approved):

txstratum/manager.py
msbrogli: Done. And I also added a new unittest. Thanks!

PR #9, opened 2021-01-15
fix: Estimate tx-mining-time when the total hashrate of the miners is unknown

msbrogli:
When a miner connects to the tx mining service, its hashrate is unknown and it will be discovered after a few iterations in the difficulty adjustment loop. Usually it takes ~30 seconds to find a reasonable estimate for the hashrate.

Currently, when the total hashrate of the miners is unknown, we set the `job.expected_mining_time` to `-1`.

This pull request intends to fix a problem that happens when several transactions are arriving to be mined and we restart the tx mining service. In this case, all miners will reconnect and their hashrate will be unknown. At the same time, several transactions will arrive and will have `expected_mining_time = -1`. In this case, the wallets will show an error message saying that "No miners are available" while the tx mining service will actually solve these transactions.

Besides that, as transactions have priority over blocks, miners will receive tx jobs (instead of block jobs) and the difficulty adjustment loop skip tx jobs. It brings a situation in which we never discover the hashrate of the miners and hence we cannot estimate the mining time of new transactions. As users keep receiving the "No miners are available" error message, they usually try again and we get the same transaction over and over.

The proposed change is currently in use by the tx mining service connected to our mainnet. It sets a fixed estimate for the mining time of transactions when the hashrate is unknown. So, the wallets won't show the error message and will work correctly. After the queue is empty, the tx mining service will send block jobs to the miners and the discovery process will work correctly.

### DIFF
--- a/txstratum/constants.py
+++ b/txstratum/constants.py
@@ -1,0 +1,18 @@
+# Copyright (c) Hathor Labs and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Default expected mining time when there is no miners or the total hashrate is unknown.
+#
+# We used to return `-1`. In this case, the wallets show an error message saying that
+# "No miners are available".
+#
+# We had an issue when there was a high demand for transaction mining and we restarted
+# the tx-mining-service. In this case, all miners will reconnect and their hashrate will
+# be unknown. At the same time, several transactions will arrive and will have expected_mining_time = -1.
+# In this case, the wallets will show an error message saying that "No miners are available"
+# while the tx mining service will actually solve these transactions.
+#
+# To fix this issue, we set a positive value for the default expected mining time.
+DEFAULT_EXPECTED_MINING_TIME: float = 5.0
--- a/txstratum/utils.py
+++ b/txstratum/utils.py
@@ -11,6 +11,8 @@ from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, Generic, Optio
 
 from structlog import get_logger
 
+from txstratum.constants import DEFAULT_EXPECTED_MINING_TIME
+
 if TYPE_CHECKING:
     from txstratum.commons import BaseTransaction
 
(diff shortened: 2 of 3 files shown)


jansegre (approved):

txstratum/utils.py
```diff
@@ -317,7 +317,7 @@ def calculate_expected_mining_time(miners_hashrate_ghs: float, job_weight: float
       5  | 0.9932
     """
     if miners_hashrate_ghs == 0:
-        return -1
+        return 5
```
pedroferreira1, line 320: Again just think we should add a comment here explaining the reason behind this change. Maybe use part of the PR description.

pedroferreira1 (approved):

txstratum/utils.py
msbrogli: Done. Can you review again, please?

pedroferreira1 (approved):

txstratum/utils.py
pedroferreira1: It's great!

txstratum/constants.py
```diff
@@ -0,0 +1,18 @@
...
+# "No miners are available".
+#
+# We had an issue when there was a high demand for transaction mining and we restarted
+# the tx-mining-service. In this case, all miners will reconnect and their hashrate will
+# be unknown. At the same time, several transactions will arrive and will have expected_mining_time = -1.
+# In this case, the wallets will show an error message saying that "No miners are available"
+# while the tx mining service will actually solve these transactions.
+#
+# To fix this issue, we set a positive value for the default expected mining time.
+DEFAULT_EXPECTED_MINING_TIME: float = 5
```
jansegre, line 18: `5` is `int`, shouldn't we use `5.0`? I'd imagine the linter would complain. But this is really minor and shouldn't cause problems on Python 3 (it used to be a bigger problem in Python 2).

jansegre (approved):

txstratum/constants.py
msbrogli: Fixed. Thanks!

PR #10, opened 2021-01-15
feat(backend): Backend's get_block_template must give an address for the block

msbrogli:
It doesn't make sense to mine blocks without an output, and to generate an output we need an address. Even though it is very unlikely that the tx-mining-service will find a block in Hathor's mainnet, it is also used to mine blocks in the testnet where it indeed find many blocks.

### DIFF
--- a/txstratum/cli.py
+++ b/txstratum/cli.py
@@ -26,6 +26,7 @@ def create_parser() -> ArgumentParser:
     parser.add_argument('--tx-timeout', help='Tx mining timeout (seconds)', type=int, default=None)
     parser.add_argument('--prometheus', help='Path to export metrics for Prometheus', type=str, default=None)
     parser.add_argument('--testnet', action='store_true', help='Use testnet config parameters')
+    parser.add_argument('--address', help='Mining address for blocks', type=str, default=None)
     parser.add_argument('backend', help='Endpoint of the Hathor API (without version)', type=str)
     return parser
 
@@ -52,7 +53,8 @@ def execute(args: Namespace) -> None:
 
     backend = HathorClient(args.backend)
     manager = TxMiningManager(
-        backend=backend
+        backend=backend,
+        address=args.address,
     )
     loop.run_until_complete(backend.start())
     loop.run_until_complete(manager.start())
--- a/txstratum/commons/client.py
+++ b/txstratum/commons/client.py
@@ -73,10 +73,13 @@ class HathorClient:
             major, minor, patch = ver.split('.')
             return HathorVersion(int(major), int(minor), int(patch))
 
-    async def get_block_template(self) -> BlockTemplate:
+    async def get_block_template(self, address: Optional[str] = None) -> BlockTemplate:
         """Return a block template."""
         assert self._session is not None
-        async with self._session.get(self._get_url('get_block_template')) as resp:
+        params = {}
+        if address is not None:
+            params['address'] = address
+        async with self._session.get(self._get_url('get_block_template'), params=params) as resp:
             data = await resp.json()
 
             if data.get('error'):
--- a/txstratum/manager.py
+++ b/txstratum/manager.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Deque, Dict, List, Optional
 
 from structlog import get_logger
 
+from txstratum.commons.utils import decode_address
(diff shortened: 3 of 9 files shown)


txstratum/cli.py
```diff
@@ -52,7 +53,8 @@ def execute(args: Namespace) -> None:
 
     backend = HathorClient(args.backend)
     manager = TxMiningManager(
-        backend=backend
+        backend=backend,
+        address=args.address,
```
pedroferreira1, line 57: Should we validate the address string? I think in the future we should have a method on hathor-commons to validate depending on the network and then this would be easy to do.

pedroferreira1 (approved):

jansegre:
The motivation for this change is unclear, `address` is in fact an optional parameter, the miner may manually add a script after the fact. I agree we should be mining blocks we can claim the reward for, but the hashrate we use is so small it doesn't make any difference.

msbrogli:
> The motivation for this change is unclear to me, `address` is in fact an optional parameter, the miner may manually add a script after the fact. I agree we should be mining blocks we can claim the reward for, but the hashrate we use is so small it doesn't make any difference.

I agree that it is optional. There was a hot fix in the production server with the comment `# XXX: API error without address`. Anyway, it makes sense to run only one miner for both blocks and transactions in a testnet. In this case, it is relevant to be able to choose the address.

txstratum/cli.py
msbrogli: I agree. I'll make this change.

jansegre:
> I agree that it is optional. There was a hot fix in the production server with the comment `# XXX: API error without address`. Anyway, it makes sense to run only one miner for both blocks and transactions in a testnet. In this case, it is relevant to be able to choose the address. @jansegre

I think that hotfix was because of a regression on the node API that broke that endpoint when no address was given. That regression is fixed and tested. But that makes sense if the testnet relies on this.

txstratum/cli.py
msbrogli: Done!

txstratum/commons/client.py
```diff
@@ -73,10 +73,13 @@ async def version(self) -> HathorVersion:
...
             return HathorVersion(int(major), int(minor), int(patch))
 
-    async def get_block_template(self) -> BlockTemplate:
+    async def get_block_template(self, address: str) -> BlockTemplate:
         """Return a block template."""
         assert self._session is not None
-        async with self._session.get(self._get_url('get_block_template')) as resp:
+        params = {}
+        if address:
+            params['address'] = address
```
jansegre, line 81: Because the address is not "optional" according to the signature, I'd rather not allow it to be empty or alternatively use `address: Optional[str] = None` on the signature and only omit when `address is None`. Either way I'd prefer is `address = ''` is invalid because it's harder to know if the address is intentionally empty or somewhere up the call stack retrieving the address failed and wasn't treated properly (like an empty env var or cli arg).

txstratum/manager.py
```diff
@@ -37,10 +38,15 @@ class TxMiningManager:
     BLOCK_TEMPLATE_UPDATE_INTERVAL = 3.0  # seconds
     TX_CLEAN_UP_INTERVAL = 300.0  # seconds
 
-    def __init__(self, backend: 'HathorClient'):
+    def __init__(self, backend: 'HathorClient', address: str):
         """Init TxMiningManager with backend."""
+        if address:
+            # Validate address. If the address is invalid, it raises an InvalidAddress exception.
+            decode_address(address)
```
jansegre, line 45: My previous comment goes for here as well.

txstratum/commons/client.py
msbrogli: Good point. I'll do it.

jansegre (approved):

txstratum/manager.py
msbrogli: Fixed. Can you review again, please?

txstratum/commons/client.py
msbrogli: Fixed. Can you review again, please?

jansegre (approved):

txstratum/commons/client.py
msbrogli: You told me it's good in Slack.

txstratum/manager.py
msbrogli: You told me it's good in Slack.

pedroferreira1 (approved):

PR #17, opened 2021-01-20
fix: Create a txstratum.time.time() function to get current timestamp

msbrogli:


jansegre (approved):

tests/utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Clock:
+    def __init__(self, loop: Optional[AbstractEventLoop]):
+        if loop is None:
+            loop = asyncio.get_event_loop()
+        self.loop = loop
+        self.ref_time = txstratum.time.time()
+
+    def time(self):
```
pedroferreira1, line 22: Missing return type

pedroferreira1 (approved):

tests/utils.py
msbrogli: Fixed. Thanks!

PR #21, opened 2021-03-04
refactor: Use hathorlib

msbrogli:


pedroferreira1 (approved):

pyproject.toml
```diff
@@ -23,6 +23,7 @@ structlog = "^20.2.0"
 prometheus-client = "^0.9.0"
 idna_ssl = "^1.1.0"
 asynctest = "^0.13.0"
+hathorlib = {version = "^0.0.3", extras = ["client"]}
```
jansegre, line 26: I think we can do a proper initial release as `v0.1.0` for hathorlib, since everything seems to be working now.

jansegre (approved):

pyproject.toml
msbrogli: Even though it is working, I feel we need a better documentation and initialization before releasing `v0.1.0`. As this version is working fine with this service, I'd just keep this version for now. What do you think?

jansegre: OK, no problem.

PR #22, opened 2021-03-04
test: Add GitHub Workflow

msbrogli:


.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install Poetry
+      run: pip -q --no-input install poetry
+    - name: Install Poetry dependencies
+      run: poetry install -n -E client --no-root
```
jansegre, line 34: `-E client` is for adding the `client` extras, I added this on `hathorlib` since it has that extra, I don't think it should be here

jansegre (approved):

.github/workflows/main.yml
msbrogli: Fixed. Thanks!

PR #24, opened 2021-03-05
feat: Add support to job resubmit in case of timeout

msbrogli:


tests/test_manager.py
```diff
@@ -669,3 +669,22 @@ async def test_block_timestamp_update(self):
         await self.advance(40)
         job.update_timestamp()
         self.assertEqual(int(old_ts), job._block.timestamp)
+
+    async def test_tx_resubmit(self):
+        job1 = TxJob(TX1_DATA, timeout=10)
+        ret1 = self.manager.add_job(job1)
+        self.assertTrue(ret1)
+
+        # When the a similar job is submitted, manager declines it.
```
jansegre, line 678: "When there is a"?

jansegre (approved):

tests/test_manager.py
msbrogli: Fixed. Thanks!

pedroferreira1: I think the 'the' should be delete.

pedroferreira1 (approved):

pedroferreira1 (approved):

jansegre (approved):

PR #33, opened 2021-09-24
chore: add Dockerfile

luislhl:
closes: https://github.com/HathorNetwork/tx-mining-service/issues/29

### Acceptance Criteria
- It should be possible to run the service using Docker
- The Docker image should be uploaded to Docker Hub when releases are created in Github

### Notes
I tested the upload using a fork of the repo and my personal account in Docker Hub.

The [release](https://github.com/luislhl/tx-mining-service/releases/tag/v0.0.1) correctly triggered the upload to https://hub.docker.com/r/luislhl/tx-mining-service/tags

### TODO
- [x] Create the repo in Docker Hub and configure the secrets in this repo


Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# copy the dependencies file to the working directory
+COPY poetry.lock pyproject.toml /code/
+
+# install dependencies
+RUN poetry config virtualenvs.create false \
+  && poetry install --no-dev --no-interaction --no-ansi
+
+# copy the content of the local src directory to the working directory
+COPY . .
```
jansegre, line 18: This probably isn't great without a `.dockerignore`, but I'd rather have it copy what's actually needed, I think only `txstratum`. Still would be better to ignore `*.py[oc]` that might be inside `txstratum`.

jansegre (changes requested):
I think this is a good start, but it would be great to make the CI build and push images. And there's also a lot of optimizations to image size and build times that can be borrowed from `hathor-core` that I can help with.

Dockerfile
luislhl: Done!

luislhl:
> I think this is a good start, but it would be great to make the CI build and push images. And there's also a lot of optimizations to image size and build times that can be borrowed from `hathor-core` that I can help with.

@jansegre , I've implemented some optimizations on image size. It's now 97 MB, down from 201 MB.

Do you see any further optimization?

I'm still working in the CI build.

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /code
+
+RUN pip install --no-cache-dir 'poetry==1.1.6'
+
+COPY poetry.lock pyproject.toml /code/
+
+RUN poetry config virtualenvs.create false \
+  && poetry install --no-dev --no-interaction --no-ansi
+
+FROM python:3.8-alpine
```
r4mmer, line 12: In this case wouldn't it be better to use the same image on the build and final stages?

README.md
```diff
@@ -18,6 +18,11 @@ To install dependencies, run `poetry install`.
 
     python main.py https://node1.mainnet.hathor.network/
 
+Or you can use the Docker image
+
+    docker build -t tx-mining-service .
+    docker run -it --network="host" tx-mining-service <full-node-address:port> --address <mining-reward-address> --api-port 8080 --stratum-port 8000
```
r4mmer, line 24: Since we are pushing to a public registry we could just skip the local build and use the public image.

r4mmer (changes requested):

README.md
luislhl: True. But maybe someone wants to make changes and run his/her customized bersion in Docker.

I think it's useful to have the instructions to build locally.

But I can add a line about using the public image.

Dockerfile
luislhl: Yeah, I was in doubt if this would generate some problem, but apparently everything worked.
I did this to make the image smaller.

Using the `alpine` in the build stage didn't work, we get compilation errors.

Since everything worked, I thought this was a safe choice.

luislhl: But I can spend some more time trying to figure out if we can fix the compilation erros in alpine, which I didn't do yet.

luislhl: Done, alpine just needed some additional dependencies to be able to compile things

README.md
luislhl: ADded

r4mmer (approved):

jansegre (approved):

PR #34, opened 2021-11-04
feat(api): Add support for banned tx ids and addresses

msbrogli:


jansegre (approved):
Although using a static file that is never reloaded is very limited and should be outdated very soon. This is the de facto master code and should be useful for the next set of changes that introduce dynamically checking the ban list.

r4mmer (approved):

PR #35, opened 2021-11-04
feat(api): Add toi client

r4mmer:
# Acceptance criteria

Create a client for toi service

### DIFF
--- a/txstratum/api.py
+++ b/txstratum/api.py
@@ -18,6 +18,7 @@ from txstratum.utils import tx_or_block_from_bytes
 
 if TYPE_CHECKING:
     from txstratum.manager import TxMiningManager
+    from txstratum.toi_client import TOIAsyncClient
 
 
 # Default maximum txout script size (in bytes).
@@ -42,7 +43,7 @@ class App:
     def __init__(self, manager: 'TxMiningManager', *, max_tx_weight: Optional[float] = None,
                  max_timestamp_delta: Optional[int] = None, tx_timeout: Optional[float] = None,
                  fix_invalid_timestamp: bool = False, max_output_script_size: Optional[int] = None,
-                 only_standard_script: bool = True):
+                 only_standard_script: bool = True, toiclient: Optional['TOIAsyncClient'] = None):
         """Init App."""
         super().__init__()
         self.log = logger.new()
@@ -52,8 +53,7 @@ class App:
         self.max_timestamp_delta: float = max_timestamp_delta or MAX_TIMESTAMP_DELTA
         self.tx_timeout: float = tx_timeout or TX_TIMEOUT
         self.only_standard_script: bool = only_standard_script
-        self.banned_tx_ids: Set[bytes] = set()
-        self.banned_addresses: Set[str] = set()
+        self.toiclient = toiclient
         self.app = web.Application()
         self.app.router.add_get('/health-check', self.health_check)
         self.app.router.add_get('/mining-status', self.mining_status)
@@ -112,20 +112,23 @@ class App:
             self.log.debug('non-standard-tx', data=data)
             return web.json_response({'error': 'non-standard-tx'}, status=400)
 
-        # Check for banned inputs.
-        for txin in tx.inputs:
-            if txin.tx_id in self.banned_tx_ids:
-                self.log.info('txin-banned', data=data)
-                return web.json_response({'error': 'invalid-tx'}, status=400)
+        if self.toiclient:
+            # Check for banned txs and addresses.
+            txs: Set[str] = set()
+            addrs: Set[str] = set()
+            for txin in tx.inputs:
+                txs.add(txin.tx_id.decode())
 
-        # Check for banned addresses.
-        for txout in tx.outputs:
-            p2pkh = P2PKH.parse_script(txout.script)
-            if p2pkh is not None:
-                self.log.info('p2pkh.address', address=p2pkh.address)
-                if p2pkh.address in self.banned_addresses:
-                    self.log.info('banned-address', data=data)
-                    return web.json_response({'error': 'invalid-tx'}, status=400)
+            for txout in tx.outputs:
+                p2pkh = P2PKH.parse_script(txout.script)
+                if p2pkh is not None:
+                    self.log.info('p2pkh.address', address=p2pkh.address)
+                    addrs.add(p2pkh.address)
+
+            toiresp = await self.toiclient.check_blacklist(tx_ids=list(txs), addresses=list(addrs))
+            if toiresp.blacklisted:
+                self.log.info('banned', data=data, issues=toiresp.issues)
+                return web.json_response({'error': 'invalid-tx'}, status=400)
 
         now = txstratum.time.time()
         if abs(tx.timestamp - now) > self.max_timestamp_delta:
--- a/txstratum/api.py
+++ b/txstratum/api.py
@@ -112,7 +112,7 @@ class App:
             return web.json_response({'error': 'non-standard-tx'}, status=400)
 
         for tx_filter in self.tx_filters:
-            if tx_filter.check_tx(tx, data):
+            if await tx_filter.check_tx(tx, data):
                 return web.json_response({'error': 'invalid-tx'}, status=400)
 
         now = txstratum.time.time()
--- a/txstratum/api.py
+++ b/txstratum/api.py
@@ -4,12 +4,11 @@
 # LICENSE file in the root directory of this source tree.
 
 import json
-from typing import TYPE_CHECKING, Optional, Set
+from typing import TYPE_CHECKING, List, Optional
 
 from aiohttp import web
 from hathorlib import TokenCreationTransaction, Transaction
 from hathorlib.exceptions import TxValidationError
-from hathorlib.scripts import P2PKH
 from structlog import get_logger
 
 import txstratum.time
@@ -17,8 +16,8 @@ from txstratum.jobs import JobStatus, TxJob
 from txstratum.utils import tx_or_block_from_bytes
 
 if TYPE_CHECKING:
+    from txstratum.filters import TXFilter
     from txstratum.manager import TxMiningManager
-    from txstratum.toi_client import TOIAsyncClient
 
 
 # Default maximum txout script size (in bytes).
@@ -43,7 +42,7 @@ class App:
     def __init__(self, manager: 'TxMiningManager', *, max_tx_weight: Optional[float] = None,
                  max_timestamp_delta: Optional[int] = None, tx_timeout: Optional[float] = None,
                  fix_invalid_timestamp: bool = False, max_output_script_size: Optional[int] = None,
-                 only_standard_script: bool = True, toiclient: Optional['TOIAsyncClient'] = None):
+                 only_standard_script: bool = True, tx_filters: Optional[List['TXFilter']] = None):
         """Init App."""
         super().__init__()
         self.log = logger.new()
@@ -53,7 +52,7 @@ class App:
         self.max_timestamp_delta: float = max_timestamp_delta or MAX_TIMESTAMP_DELTA
         self.tx_timeout: float = tx_timeout or TX_TIMEOUT
         self.only_standard_script: bool = only_standard_script
-        self.toiclient = toiclient
+        self.tx_filters = tx_filters or []
         self.app = web.Application()
         self.app.router.add_get('/health-check', self.health_check)
         self.app.router.add_get('/mining-status', self.mining_status)
@@ -112,22 +111,8 @@ class App:
             self.log.debug('non-standard-tx', data=data)
             return web.json_response({'error': 'non-standard-tx'}, status=400)
 
-        if self.toiclient:
-            # Check for banned txs and addresses.
-            txs: Set[str] = set()
-            addrs: Set[str] = set()
-            for txin in tx.inputs:
-                txs.add(txin.tx_id.hex())
-
-            for txout in tx.outputs:
-                p2pkh = P2PKH.parse_script(txout.script)
-                if p2pkh is not None:
-                    self.log.debug('p2pkh.address', address=p2pkh.address)
-                    addrs.add(p2pkh.address)
-
-            toiresp = await self.toiclient.check_blacklist(tx_ids=list(txs), addresses=list(addrs))
-            if toiresp.blacklisted:
-                self.log.info('banned', data=data, issues=toiresp.issues)
+        for tx_filter in self.tx_filters:
+            if tx_filter.check_tx(tx, data):
                 return web.json_response({'error': 'invalid-tx'}, status=400)
 
         now = txstratum.time.time()
--- a/txstratum/api.py
+++ b/txstratum/api.py
@@ -117,12 +117,12 @@ class App:
             txs: Set[str] = set()
             addrs: Set[str] = set()
             for txin in tx.inputs:
-                txs.add(txin.tx_id.decode())
+                txs.add(txin.tx_id.hex())
 
             for txout in tx.outputs:
                 p2pkh = P2PKH.parse_script(txout.script)
                 if p2pkh is not None:
-                    self.log.info('p2pkh.address', address=p2pkh.address)
+                    self.log.debug('p2pkh.address', address=p2pkh.address)
                     addrs.add(p2pkh.address)
 
             toiresp = await self.toiclient.check_blacklist(tx_ids=list(txs), addresses=list(addrs))
--- a/txstratum/cli.py
+++ b/txstratum/cli.py
@@ -7,7 +7,7 @@ import logging
 import logging.config
 import os
 from argparse import ArgumentParser, Namespace
-from typing import List, Optional, Set
+from typing import List, Optional
 
 import structlog
 from aiohttp import web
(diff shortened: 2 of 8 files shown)


txstratum/api.py
```diff
@@ -112,21 +112,24 @@ async def submit_job(self, request: web.Request) -> web.Response:
...
+            for txin in tx.inputs:
+                txs.add(txin.tx_id.decode())
+
+            for txout in tx.outputs:
+                p2pkh = P2PKH.parse_script(txout.script)
+                if p2pkh is not None:
+                    self.log.info('p2pkh.address', address=p2pkh.address)
+                    addrs.add(p2pkh.address)
+
+            toiresp = await self.toiclient.check_blacklist(tx_ids=list(txs), addresses=list(addrs))
```
pedroferreira1, line 128: If ToI service is down for some reason we will block all mining for transactions?

txstratum/api.py
```diff
@@ -112,21 +112,24 @@ async def submit_job(self, request: web.Request) -> web.Response:
...
+            # Check for banned txs and addresses.
+            txs: Set[str] = set()
+            addrs: Set[str] = set()
+            for txin in tx.inputs:
+                txs.add(txin.tx_id.decode())
+
+            for txout in tx.outputs:
+                p2pkh = P2PKH.parse_script(txout.script)
+                if p2pkh is not None:
+                    self.log.info('p2pkh.address', address=p2pkh.address)
```
pedroferreira1, line 125: Maybe this should be a debug log

txstratum/api.py
```diff
@@ -112,21 +112,24 @@ async def submit_job(self, request: web.Request) -> web.Response:
...
-        # Check for banned inputs.
-        for txin in tx.inputs:
-            if txin.tx_id in self.banned_tx_ids:
-                self.log.info('txin-banned', data=data)
+        if self.toiclient:
+            # Check for banned txs and addresses.
+            txs: Set[str] = set()
+            addrs: Set[str] = set()
+            for txin in tx.inputs:
+                txs.add(txin.tx_id.decode())
```
pedroferreira1, line 120: `txin.tx_id.decode()` is working fine? At first I thought we should do `txin.tx_id.hex()`

txstratum/toi_client.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(
+            self,
+            base_url: str,
+            apikey: Optional[str] = None,
+            session: Optional[ClientSession] = None,
+    ) -> None:
+        """Configure TOIAsyncClient session and auth.
+
+        base_url: TOI Service host
+        apikey: TOI Service apikey for auth (optional)
```
pedroferreira1, line 31: Isn't it required for ToI login?

txstratum/api.py
pedroferreira1: For what I've seen it will return `blacklisted=False`, right? Anyway, I think we should have a short timeout (maybe 1s or less) otherwise we could increase the send transaction time if this service is fully loaded.

We should also have an alert if we have many timeouts in order to check why the service can't reply the requests in less than the timeout.

jansegre: I prefer having a retry mechanism (if the client using the tx-mining-service has some more room for waiting) and fallback to blocking if it fails to get a response from the toi-service.

txstratum/toi_client.py
```diff
@@ -0,0 +1,80 @@
+from dataclasses import dataclass
```
jansegre, line 1: I like this client, and it seems to be completely independent from the tx-mining-service. It looks like a good candidate to be its own lib that other services could use.

txstratum/api.py
jansegre: It shouldn't work a lot of the times (not every byte sequence is a valid str, especially when using decode without arguments, which will default to utf-8 which has more precise rules).

I'd prefer to either use a `Set[bytes]`, or @pedroferreira1's suggestion of using `txin.tx_id.hex()`.

jansegre (changes requested):

txstratum/toi_client.py
r4mmer: I'm considering this, but we should have 2 clients an async and a sync one (using aiohttp and requests). For now I think we should focus on making the toi-service operational and later we can remove this local client for a lib.

txstratum/api.py
r4mmer: Checked and the correct one should `.hex()`

txstratum/api.py
r4mmer: Done.

txstratum/api.py
jansegre: The opposite makes more sense to me because our capacity to _allow_ a transaction would depend on how resilient our services are. Which is how it naturally is. Otherwise in a chaotic high load situation, some transactions fails (because of overload) and some are allowed, but the ones that are allowed maybe should have been blocked, but weren't because we couldn't reach the toi-service.

What do you think @r4mmer?

pedroferreira1: > A user that is blacklisted can try a DoS to bypass the blockage.

I think I agree with @jansegre in this situation, it makes sense to me given the discussion that we could block by default and it's extremely important that our alert system is used in order to understand if mining is being blocked because the service is taking too long to reply or if it's receiving many requests.

luislhl: Just a comment about the alert system that was mentioned. If we want to send alerts, all that's needed is to send messages in the right format to a SQS queue.

I can help with this if needed.

msbrogli (changes requested):

txstratum/cli.py
```diff
@@ -66,29 +69,23 @@ def execute(args: Namespace) -> None:
...
     loop.run_until_complete(manager.start())
     server = loop.run_until_complete(loop.create_server(manager, '0.0.0.0', args.stratum_port))
+    toiclient: Optional[TOIAsyncClient] = None
 
     if args.prometheus:
         from txstratum.prometheus import PrometheusExporter
         metrics = PrometheusExporter(manager, args.prometheus)
         metrics.start()
 
+    if args.toi_url and args.toi_apikey:
```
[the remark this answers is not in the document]
r4mmer: Done.

txstratum/cli.py
```diff
@@ -66,29 +69,23 @@ def execute(args: Namespace) -> None:
...
+
     api_app = App(manager, max_tx_weight=args.max_tx_weight, max_timestamp_delta=args.max_timestamp_delta,
                   tx_timeout=args.tx_timeout, fix_invalid_timestamp=args.fix_invalid_timestamp,
-                  only_standard_script=not args.allow_non_standard_script)
+                  only_standard_script=not args.allow_non_standard_script, toiclient=toiclient)
     logger.info('API Configuration', max_tx_weight=api_app.max_tx_weight, tx_timeout=api_app.tx_timeout,
                 max_timestamp_delta=api_app.max_timestamp_delta, fix_invalid_timestamp=api_app.fix_invalid_timestamp,
                 only_standard_script=api_app.only_standard_script)
 
-    if args.ban_tx_ids:
```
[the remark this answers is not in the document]
r4mmer: Made a `FileFilter` for this.

txstratum/api.py
```diff
@@ -112,21 +112,24 @@ async def submit_job(self, request: web.Request) -> web.Response:
...
+                if p2pkh is not None:
+                    self.log.debug('p2pkh.address', address=p2pkh.address)
+                    addrs.add(p2pkh.address)
+
+            toiresp = await self.toiclient.check_blacklist(tx_ids=list(txs), addresses=list(addrs))
+            if toiresp.blacklisted:
+                self.log.info('banned', data=data, issues=toiresp.issues)
                 return web.json_response({'error': 'invalid-tx'}, status=400)
 
-        # Check for banned addresses.
```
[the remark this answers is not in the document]
r4mmer: Done.

txstratum/api.py
```diff
@@ -112,21 +111,10 @@ async def submit_job(self, request: web.Request) -> web.Response:
             self.log.debug('non-standard-tx', data=data)
             return web.json_response({'error': 'non-standard-tx'}, status=400)
 
-        # Check for banned inputs.
-        for txin in tx.inputs:
-            if txin.tx_id in self.banned_tx_ids:
-                self.log.info('txin-banned', data=data)
+        for tx_filter in self.tx_filters:
+            if await tx_filter.check_tx(tx, data):
                 return web.json_response({'error': 'invalid-tx'}, status=400)
```
msbrogli, line 116: Should we let the filter decides that return?

txstratum/cli.py
```diff
@@ -66,29 +72,52 @@ def execute(args: Namespace) -> None:
...
+            raise ValueError("Should pass both toi_url and toi_apikey")
+        toiclient = TOIAsyncClient(args.toi_url, args.toi_apikey)
+        tx_filters.append(TOIFilter(toiclient))
+
+    api_app = App(manager, max_tx_weight=args.max_tx_weight, max_timestamp_delta=args.max_timestamp_delta,
+                  tx_timeout=args.tx_timeout, fix_invalid_timestamp=args.fix_invalid_timestamp,
+                  only_standard_script=not args.allow_non_standard_script, tx_filters=tx_filters)
+    logger.info('API Configuration', max_tx_weight=api_app.max_tx_weight, tx_timeout=api_app.tx_timeout,
+                max_timestamp_delta=api_app.max_timestamp_delta, fix_invalid_timestamp=api_app.fix_invalid_timestamp,
+                only_standard_script=api_app.only_standard_script)
```
msbrogli, line 119: Should we log which filters are in use?

txstratum/cli.py
```diff
@@ -66,29 +72,52 @@ def execute(args: Namespace) -> None:
...
             line = line.strip()
             if not line:
                 continue
             logger.info('Added to banned tx ids', txid=line)
-            api_app.banned_tx_ids.add(bytes.fromhex(line))
+            banned_tx_ids.add(bytes.fromhex(line))
+        fp.close()
+
+    if args.ban_addrs:
+        fp = open(args.ban_addrs, 'r')
```
msbrogli, line 96: Why don't we move it to the `FileFilter`? Maybe add a method `load_from_file(path: str)`?

txstratum/cli.py
```diff
@@ -103,6 +132,8 @@ def execute(args: Namespace) -> None:
     except KeyboardInterrupt:
         logger.info('Stopping...')
 
+    if toiclient is not None:
```
msbrogli, line 135: Maybe we should do it for all filters.

txstratum/filters.py
```diff
@@ -0,0 +1,81 @@
...
+        raise NotImplementedError
+
+
+class FileFilter(TXFilter):
+    """Filter tx based on a set of banned tx_ids and addresses."""
+
+    def __init__(self, banned_tx_ids: Set[bytes] = set(), banned_addresses: Set[str] = set()) -> None:
+        """Init filter."""
+        self.log = logger.new()
+        self.banned_tx_ids = banned_tx_ids
```
msbrogli, line 36: It seems like a static filter. I feel it should receive a file and read from there. We can even detect a file change and update this list automatically without a restart (or we can do it in the future).

txstratum/filters.py
```diff
@@ -0,0 +1,81 @@
...
+        self.banned_addresses = banned_addresses
+
+    async def check_tx(self, tx: 'BaseTransaction', data: Any) -> bool:
+        """Return if the tx should be blocked."""
+        if len(self.banned_tx_ids):
+            for txin in tx.inputs:
+                if txin.tx_id in self.banned_tx_ids:
+                    return True
```
msbrogli, line 44: Maybe add a log message?

txstratum/toi_client.py
```diff
@@ -0,0 +1,85 @@
...
+            if response.status == 200:
+                r = await response.json()
+                return CheckBlacklist(**r)
+            self.log.warning(
+                message='TOI Service unexpected response',
+                status=response.status,
+                body=await response.text(),
+            )
+            # XXX: Allow if service fails call
+            return CheckBlacklist(blacklisted=False)
```
msbrogli, line 85: I feel the client should not decide the final behavior. Just raise an exception and let the filter decides.

msbrogli (changes requested):

txstratum/api.py
msbrogli: I think Jan made good considerations about it. I feel we should have a way to enable/disabled a circuit breaker as needed. This might be a flag in a Redis, or just an administrative API to enable/disable the circuit breaker. For clarity, all transactions will be accepted by the TOI Filter when the circuit breaker is enabled.

Here are a few other ideas to discuss:

1) Should we keep track of the last N requests to the TOI Service and use it in case of a failure? We can use an internal redis for this so we can safely restart the tx-mining-service daemon.

2) Maybe we should add a metric of failure from the TOI Cliente. We can monitor this metric and alert accordingly.

jansegre: > I feel we should have a way to enable/disabled a circuit breaker as needed. This might be a flag in a Redis, or just an administrative API to enable/disable the circuit breaker.

It makes sense to have a way to enable/disable in runtime, but I think using an external service (database or otherwise) creates the same problem again (even if the service is much less likely to fail). I'd rather it be in-memory and use an internal API to change the state in memory, while the default state would be passed either from a command line argument, environment variable, configuration file, whatever fits best (if we have other parameters just adding a new one with the same mechanism).

> Here are a few other ideas to discuss:
> 
>     1. Should we keep track of the last N requests to the TOI Service and use it in case of a failure? We can use an internal redis for this so we can safely restart the tx-mining-service daemon.

I think a local LRU cache is good and simple enough. And that would be beneficial regardless of the circuit-break enabled state.

>     2. Maybe we should add a metric of failure from the TOI Cliente. We can monitor this metric and alert accordingly.

This could be very useful to guide us to change or reinforce our decision on the default circuit-break state. If the TOI service ends up being extremely stable and not fail even when the network is under a lot of stress, it would be really support not enabling the circuit-break, otherwise if it ends up being common to have random failures, specially if under stress before the other services, than the circuit-break would be much more justifiable IMO.

What do you think?

txstratum/cli.py
r4mmer: Done.

txstratum/cli.py
r4mmer: Done.

txstratum/cli.py
r4mmer: Done.

txstratum/filters.py
r4mmer: It is a static filter. I think the idea of watching changes on a file and loading them at runtime is great, but it should be done on it's own PR.

txstratum/toi_client.py
r4mmer: Moved this decision to the filter.

txstratum/filters.py
r4mmer: Done.

txstratum/api.py
r4mmer: I don't think so, we could later have more complex filters so returning a named tuple (e.g. `(block, message)`) could be better but for now the simple boolean response indicating if the tx should be blocked is a better format.

txstratum/filters.py
```diff
@@ -0,0 +1,132 @@
...
+            resp = await self.client.check_blacklist(tx_ids=list(txs), addresses=list(addrs))
+            if resp.blacklisted:
+                self.log.info('banned', data=data, issues=resp.issues)
+            return resp.blacklisted
+        except TOIError as ex:
+            # TOI Service error
+            self.log.error("toi_service_error", data=data, body=str(ex))
+            if self.block:
+                # We are rejecting a tx based on a toi faillure
+                # TODO: alert
```
r4mmer, line 128: Including a "[ALERT]" string on the message is the old method of alerting, the new method will be implemented on #39

txstratum/filters.py
```diff
@@ -0,0 +1,132 @@
...
+    @abstractmethod
+    async def close(self) -> None:
+        """Close any open session."""
+        raise NotImplementedError
+
+
+class FileFilter(TXFilter):
+    """Filter tx based on a set of banned tx_ids and addresses."""
+
+    def __init__(self, banned_tx_ids: Set[bytes] = set(), banned_addresses: Set[str] = set()) -> None:
```
msbrogli, line 40: Maybe create a new issue to make `FileFilter` check for the file last modified date and automatically update in case of a file update.

txstratum/filters.py
```diff
@@ -0,0 +1,132 @@
...
+        return False
+
+
+class TOIFilter(TXFilter):
+    """Filter tx based on the toi service."""
+
+    def __init__(self, client: 'TOIAsyncClient', block: bool = False) -> None:
+        """Init filter."""
+        self.client = client
+        self.block = block
```
msbrogli, line 98: Add comments explaining attributes.

txstratum/toi_client.py
```diff
@@ -0,0 +1,91 @@
...
+
+from aiohttp import ClientSession
+from structlog import get_logger
+
+logger = get_logger()
+
+
+class TOIError(Exception):
+    """Raised upon any unexpected responses from toi service."""
+
```
msbrogli, line 18: Empty line?

txstratum/cli.py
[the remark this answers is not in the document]
msbrogli: Thanks!

txstratum/cli.py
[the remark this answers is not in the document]
msbrogli: Thanks!

txstratum/api.py
msbrogli: Ok. I agree. Thanks!

txstratum/cli.py
msbrogli: Thanks!

txstratum/cli.py
msbrogli: Thanks!

txstratum/cli.py
msbrogli: Thanks!

txstratum/filters.py
msbrogli: I agree. Would you open an issue with a brief description, please?

txstratum/filters.py
msbrogli: Thanks!

txstratum/toi_client.py
msbrogli: Thanks!

msbrogli (approved):

jansegre (approved):

PR #38, opened 2021-11-27
Fix: exceptions during Tx Mining

luislhl:
### Acceptance Criteria
- Fix exceptions when 2 miners submit a solution to the same job at the same time
- Fix exception when trying to timeout a job and it is not in the queue anymore

The exceptions being fixed are better described below.

### Important
Those exceptions apparently were greatly affecting the TxMiningService efficiency when receiving too many Txs in a short period of time.

They might be one of the causes for the bottlenecks described in https://github.com/HathorNetwork/internal-issues/issues/63

I ran load tests against this branch and the master to compare the effect this fix would have. The load test consisted in sending 6 txs per second for 20s in a private local network.

With the fix, 89.6% of the Txs were mined and returned to the wallet-headless as successful.

Without the fix, 40,8% were successful.

(Those load tests can be reproduced using the code in https://github.com/HathorNetwork/ops-tools/pull/221. However, a bigger setup of the environment is needed that is not described yet)

### Exceptions
This one occurred when 2 miners submitted a solution to the same job almost at the same time, and the tx job had already been popped from the queue, which them becomes empty.
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
  File "/usr/lib/python3.6/asyncio/selector_events.py", line 721, in _read_ready
    self._protocol.data_received(data)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 114, in data_received
    self.line_received(line)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 155, in line_received
    self.json_received(data)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 159, in json_received
    if self.try_as_request(data):
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 175, in try_as_request
    self.handle_request(method, params, msgid)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 208, in handle_request
    fn(params, msgid)
  File "/home/ubuntu/tx-mining-service/txstratum/protocol.py", line 230, in method_submit
    self.manager.submit_solution(self, job, obj.get_struct_nonce())
  File "/home/ubuntu/tx-mining-service/txstratum/manager.py", line 136, in submit_solution
    tx_job2 = self.tx_queue.popleft()
IndexError: pop from an empty deque
```

This is almost the same situation as the exception above, with the different that the queue is not empty. Instead, a job different from the expected is popped because the right job had already been popped by the other submission.

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
  File "/usr/lib/python3.6/asyncio/selector_events.py", line 721, in _read_ready
    self._protocol.data_received(data)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 114, in data_received
    self.line_received(line)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 155, in line_received
    self.json_received(data)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 159, in json_received
    if self.try_as_request(data):
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 175, in try_as_request
    self.handle_request(method, params, msgid)
  File "/home/ubuntu/tx-mining-service/txstratum/utils.py", line 208, in handle_request
    fn(params, msgid)
  File "/home/ubuntu/tx-mining-service/txstratum/protocol.py", line 230, in method_submit
    self.manager.submit_solution(self, job, obj.get_struct_nonce())
  File "/home/ubuntu/tx-mining-service/txstratum/manager.py", line 137, in submit_solution
    assert tx_job is tx_job2
AssertionError
```

This one happened when trying to timeout a job that was not in queue anymore. The cause is similar to the others. A miner might submit a solution just before the timeout was triggered.

```
Traceback (most recent call last):
  File "/home/luislhl/.pyenv/versions/3.8.3/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/luislhl/Workspace/hathor/tx-mining-service/txstratum/manager.py", line 267, in _job_timeout_if_possible
    self.tx_queue.remove(job)
ValueError: deque.remove(x): x not in deque
```

### DIFF
--- a/txstratum/manager.py
+++ b/txstratum/manager.py
@@ -133,8 +133,14 @@ class TxMiningManager:
             assert isinstance(job, MinerTxJob)
             # Remove from queue.
             tx_job = job.tx_job
-            tx_job2 = self.tx_queue.popleft()
-            assert tx_job is tx_job2
+            self.log.info(f"Solution submitted for job {tx_job}")
+            try:
+                self.tx_queue.remove(tx_job)
+            except ValueError:
+                # This means another miner submitted a solution shortly before
+                #  and so the txJob has been already removed from the queue
+                self.log.warning('Tried removing tx_job from the queue but the queue does not contain this tx_job. Probably because 2 miners found a solution for it at the same time.')
+                return
             # Schedule to clean it up.
             self.schedule_job_clean_up(tx_job)
             # Mark as solved, stop mining it, and propagate if requested.
@@ -264,7 +270,10 @@ class TxMiningManager:
         self.log.info('TxJob timeout', job=job.to_dict())
         self.txs_timeout += 1
         job.status = JobStatus.TIMEOUT
-        self.tx_queue.remove(job)
+        try:
+            self.tx_queue.remove(job)
+        except ValueError:
+            self.log.warning('TxJob timeout but not in queue', job=job)
         self.stop_mining_tx(job)
         # Schedule to clean it up.
         self.schedule_job_clean_up(job)
--- a/txstratum/manager.py
+++ b/txstratum/manager.py
@@ -133,7 +133,7 @@ class TxMiningManager:
             assert isinstance(job, MinerTxJob)
             # Remove from queue.
             tx_job = job.tx_job
-            self.log.info(f"Solution submitted for job {tx_job}")
+            self.log.info(f"Solution submitted for TxJob {tx_job}")
             try:
                 self.tx_queue.remove(tx_job)
             except ValueError:
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ Or you can use the Docker image
 We also have a public Docker image if you don't want to build it yourself:
     docker run -it --network="host" hathornetwork/tx-mining-service <full-node-address:port> --address <mining-reward-address> --api-port 8080 --stratum-port 8000
 
+### Debug logs
+
+If you need to enable debug logs, you should edit the file `log.conf` and set the log level there.
+
+You could also use a completely different file to configure the logs, by using the parameter `log-config` when running the service.
+
 
 ## How to test?
 


txstratum/manager.py
```diff
@@ -133,8 +133,17 @@ def submit_solution(self, protocol: StratumProtocol, job: MinerJob, nonce: bytes
...
+            self.log.info(f"Solution submitted for TxJob {tx_job}")
+            try:
+                self.tx_queue.remove(tx_job)
+            except ValueError:
+                # This means another miner submitted a solution shortly before
+                #  and so the txJob has been already removed from the queue
+                self.log.warning((
+                    "Tried removing tx_job from the queue but the queue does not contain this tx_job. "
+                    "Probably because 2 miners found a solution for it at the same time."
+                ))
```
jansegre, line 145: I think there is either a small misunderstanding or wrong (IMO) implementation. A job is supposed to be unique to a miner (if this is not the case the implementation is wrong IMO), as such it would be impossible for a different miner to submit a solution to the same job.

However it _is_ possible for a miner to submit more than one solution to the same job, in the context of the stratum protocol that makes sense, more than one `nonce` can solve the same job and if the miner has not switched jobs it will keep searching or another `nonce` on the same job, so that may actually be what is happening here. If it is implemented how I think it is, then the the "fix" IMO should be to either accept multiple solutions, or ignore it "cleanly", so maybe not with a warning, since this is expected.

txstratum/manager.py
```diff
@@ -264,7 +273,10 @@ def _job_timeout_if_possible(self, job: TxJob) -> None:
         self.log.info('TxJob timeout', job=job.to_dict())
         self.txs_timeout += 1
         job.status = JobStatus.TIMEOUT
-        self.tx_queue.remove(job)
+        try:
+            self.tx_queue.remove(job)
+        except ValueError:
+            self.log.warning('TxJob timeout but not in queue', job=job)
```
jansegre, line 279: IMO the correct implementation wouldn't necessarily _remove_ the job when it is submitted. If this is the case, then jobs would _only_ be removed by timeout, and then it would really make sense to expect or at least log an error when the jobs is not (because it wouldn't be possible for it to have been removed in other way).

txstratum/manager.py
luislhl: > as such it would be impossible for a different miner to submit a solution to the same job.

I've run a test to validate this, adding a log to show which miner solved which job. Apparently different miners are indeed submitting solutions to the same job.

Those are the logs I got:

```
txstratum.manager - INFO - 2021-12-01 11:46.36 [info     ] Solution submitted for TxJob <txstratum.jobs.TxJob object at 0x7f17be5f0340> by miner 39a1cb92-c919-4676-88d2-e8c2ae28a9fd
txstratum.manager - INFO - 2021-12-01 11:46.36 [info     ] TxJob solved                   propagate=False tx_job={'job_id': 'e3378b40fae08f7a979d211907ecb86b619ee49379b8f304d6368c4c41da4128', 'status': 'done', 'message': '', 'created_at': 1638369994.900104, 'tx': {'nonce': '000eaeba', 'parents': ['000026147de2f75ff1d645e28f52a1391830b42fc7d9894c594c8972d5b2a4d3', '000000e0f6eb02a681596c8add13efea019bb508856e82494785882309d1d10d'], 'timestamp': 1638369995, 'weight': 19.380205270799852}, 'timeout': 20.0, 'submitted_at': 1638369996.085033, 'total_time': 1.1849288940429688, 'expected_queue_time': 0, 'expected_mining_time': 0.10115834504104501, 'expected_total_time': 0.10115834504104501}
txstratum.manager - INFO - 2021-12-01 11:46.36 [info     ] Solution submitted for TxJob <txstratum.jobs.TxJob object at 0x7f17be5f0340> by miner c1d89a9f-11c2-4098-8262-2b0be621325c
txstratum.manager - WARNING - 2021-12-01 11:46.36 [warning  ] Tried removing tx_job from the queue but the queue does not contain this tx_job. Probably because 2 miners found a solution for it at the same time.
```

You can see that we have `Solution submitted for TxJob` 2 times for the same job, from different miners. Then the warning is logged.

> A job is supposed to be unique to a miner

I think this is true for the `MinerTxJob` instances. But they are wrappers for `TxJob` instances, and although the `MinerTxJob` instances are unique to each miner, the `TxJob` instances are not. And what we have in the queue are `TxJob` instances. That's why this behavior is happening, if I understood it correctly.

txstratum/manager.py
luislhl: But if we don't remove a job from the queue after it is submitted, won't it keep being sent to miners until the timeout is reached? Or maybe I'm not getting the whole picture you have in mind.

jansegre: I think I get it now, mostly because it's a `TxJob` instead of a `MinerTxJob`.

txstratum/manager.py
jansegre: Got it, that makes sense. Still, if this is expected and common, maybe it shouldn't be a warning. The only reason I think it makes sense for it to be a warning is because currently we can't distinguish it from a job that never existed.

jansegre (approved):

jansegre (approved):

txstratum/manager.py
```diff
@@ -264,7 +279,10 @@ def _job_timeout_if_possible(self, job: TxJob) -> None:
         self.log.info('TxJob timeout', job=job.to_dict())
         self.txs_timeout += 1
         job.status = JobStatus.TIMEOUT
-        self.tx_queue.remove(job)
+        try:
+            self.tx_queue.remove(job)
+        except ValueError:
```
pedroferreira1, line 284: If the reason for this exception really is that a solution was found right before executing the timeout (and not that maybe we had 2 timeouts scheduled for the same job), then the code below shouldn't be executed, right? In the `submit_solution` method we already execute `stop_mining_tx` and `schedule_job_clean_up`. It doesn't seem to harm because both methods are ok to be executed twice but may not be needed.

Anyway, I think we should handle this earlier in the method, otherwise we are changing the job status to the wrong status and wrongly adding 1 to the timeout count.

txstratum/manager.py
```diff
@@ -133,8 +133,23 @@ def submit_solution(self, protocol: StratumProtocol, job: MinerJob, nonce: bytes
             assert isinstance(job, MinerTxJob)
             # Remove from queue.
             tx_job = job.tx_job
-            tx_job2 = self.tx_queue.popleft()
-            assert tx_job is tx_job2
+
+            if tx_job.status == JobStatus.DONE:
```
pedroferreira1, line 137: What if the status is `TIMEOUT`, `FAILED`, or `CANCELLED`?

txstratum/manager.py
luislhl: > If the reason for this exception really is that a solution was found right before executing the timeout (and not that maybe we had 2 timeouts scheduled for the same job)

Yeah, I'm not really sure what is the cause for the exception here.

I don't think it's because of a solution found, neither because of 2 timeouts, because both conditions would be caught in the first if:
```python
        if job.status in JobStatus.get_after_mining_states():
            return
```

I think the real cause for this exception was the other exception that we are also fixing in this PR. By looking at the code where the other exception happens, you will see that it prevented all those things from running in the `submit_solution` method:

```python
            self.schedule_job_clean_up(tx_job)
            # Mark as solved, stop mining it, and propagate if requested.
            tx_job.mark_as_solved(nonce=nonce, timestamp=job.get_object().timestamp)
            self.stop_mining_tx(tx_job)
```

So, we not only would not be marked the job as `DONE`, but also we wouldn't schedule its cleaning to remove the timeout scheduled, and we would not stop other miners from mining it, which is the worse thing.

So, after this other exception happened, the timeout would still be triggered even for a tx that was solved, because it was not marked as solved. But it could have been removed from the queue. All these things would generate the exception in this timeout code.

>  then the code below shouldn't be executed, right? In the submit_solution method we already execute stop_mining_tx and schedule_job_clean_up. It doesn't seem to harm because both methods are ok to be executed twice but may not be needed.

Because of the explanation above, if things work ok, they won't be executed twice. But if something unexpected happens, we would make sure they run at least during the timeout.

> Anyway, I think we should handle this earlier in the method, otherwise we are changing the job status to the wrong status and wrongly adding 1 to the timeout count.

I think the first if in the method is protecting agaisnt this. An exception would be raised here only in strage situations. I'm thinking even in logging it as an error, or not treating this at all (just the other one).

But I guess handling this earlier doesn't hurt.

txstratum/manager.py
luislhl: I'll add the in the condition

txstratum/manager.py
luislhl: I'll change the log to an error

pedroferreira1: I agree with the explanation and the possible root cause for the problem. Maybe you should change the PR description, it makes more sense with what you commented here.

pedroferreira1: I agree with the explanation and the possible root cause for the problem. Maybe you should change the PR description, it makes more sense with what you commented here.

pedroferreira1 (approved):

jansegre (approved):

PR #40, opened 2021-12-08
feat(job): boost block weight when too low

jansegre:


txstratum/jobs.py
```diff
@@ -10,13 +10,28 @@
...
+
+# XXX: This serves to artificially boost the weight until MIN_WEIGHT_EXTRA is reached, boosting the weight is useful
+#      for making it harder for cpuminers to flood the network in case something goes wrong, it also serves to force a
+#      higher weight in case the weight is too small to meaningfully increase the block score (in the future this issue
+#      will probably have been fixed, so it won't matter much). The particular values bellow assume a hashrate of at
+#      least .5GH/s is being used on the tx-mining-service, this is 5% of the mininmum hashrate a USB GekkoScience ASIC
+#      can do, and about what 2 GPUs on AWS can do. The boosting mechanism is to basically add WEIGHT_EXTRA_ADD capping
+#      it to MIN_WEIGHT_EXTRA (examples: weight=21.0 -> weight=25.0, weight=30.0 -> weight=34.0, weight=31.0 ->
+#      weight=34.0, weight=34.0 -> weight=34.0, weight=35.0 -> weight=35.0)
+MIN_WEIGHT_EXTRA = 34.0
```
pedroferreira1, line 32: I was a bit confused with this param name at first, to be honest. I was thinking about something like `MAX_WEIGHT_TO_ADD_EXTRA`. I'm okay to move with the current name as well

pedroferreira1 (approved):

PR #41, opened 2021-12-08
feat: tweak intervals to lower pressure on the node

jansegre:
Honestly I'm not sure if we should merge this. I took this changes from the tx-mining-service in production on the testnet.


txstratum/protocol.py
```diff
@@ -40,7 +40,7 @@ class StratumProtocol(JSONRPCProtocol):
     ESTIMATOR_LOOP_INTERVAL = 15  # in seconds, "frequency" that the function that updates the estimator will be called
     ESTIMATOR_WINDOW_INTERVAL = 60 * 15  # in seconds, size of the window to use for estimating miner's hashrate
     TARGET_JOB_TIME = 15  # in seconds, adjust difficulty so jobs take this long
-    JOB_UPDATE_INTERVAL = 2  # in seconds, to update the timestamp of the job
+    JOB_UPDATE_INTERVAL = 10  # in seconds, to update the timestamp of the job
```
pedroferreira1, line 43: This is the interval to update the job? So we would wait 10s to start a job tx if the miner is with a block job?

PR #42, opened 2021-12-08
feat: accept address from wrong network

jansegre:
Although this may sound undesirable, our patches on cgminer/ccminer/cpuminer don't support using a different address per stratum server. So this is a workaround to allow a single miner to be shared on a mainnet tx-mining-service and a testnet tx-mining-service, which is exactly what I'm doing with my USB ASIC.

### DIFF
--- a/txstratum/protocol.py
+++ b/txstratum/protocol.py
@@ -8,7 +8,9 @@ from math import log2
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, NamedTuple, Optional, cast
 
 from hathorlib import sum_weights
+from hathorlib.conf import HathorSettings
 from hathorlib.exceptions import InvalidAddress
+from hathorlib.scripts import binary_to_int
 from hathorlib.utils import decode_address
 from structlog import get_logger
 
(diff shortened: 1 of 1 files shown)


luislhl (approved):
LGTM

txstratum/protocol.py
```diff
@@ -151,7 +154,13 @@ def method_subscribe(self, params: Any, msgid: JSONRPCId) -> None:
         if params and 'address' in params and params['address'] is not None:
             try:
                 address = params['address']
-                self.miner_address = decode_address(address)
+                address_bytes = decode_address(address)
```
pedroferreira1, line 157: Should we accept this only for testnet? Or maybe add a parameter to allow it, I don't know. As you said in the description, it makes sense in the context but no sense for code context.

And where do we raise `InvalidAddress` error? Inside `decode_address` method? Do we validate the network byte there?

jansegre: Honestly I think we can safely accept (logging a warning) a wrong address on the testnet for the purposes of mining blocks, tx-mining-service has the least chance to find a block on the mainnet (least chance is maybe putting it lightly, it has absolutely no chance).

jansegre: `decode_address` will raise `InvalidAddress` when it has the wrong length or when the checksum doesn't match, but it will not validate the network byte, this is only done on `create_output_script` much later.

pedroferreira1: Yeah, I agree with you

pedroferreira1 (approved):

PR #44, opened 2021-12-10
feat: Allow exposing Prometheus metrics in http port

luislhl:
### Acceptance Criteria
- We should be able to optionally enable a http server exposing Prometheus metrics in a provided port
- The other method of exporting metrics to a file should still be allowed

### Motivation
Depending on the infrastructure where the service is deployed, it may be better to collect the metrics through a http endpoint (it's actually the usual method in Prometheus)


txstratum/cli.py
```diff
@@ -28,6 +28,7 @@ def create_parser() -> ArgumentParser:
     parser.add_argument('--tx-timeout', help='Tx mining timeout (seconds)', type=int, default=None)
     parser.add_argument('--fix-invalid-timestamp', action='store_true', help='Fix invalid timestamp to current time')
     parser.add_argument('--prometheus', help='Path to export metrics for Prometheus', type=str, default=None)
+    parser.add_argument('--prometheus-port', help='Enables exporting metrics in a http port', type=int, default=9090)
```
pedroferreira1, line 31: If we have this default value, then how can this be optional? If I don't pass the `--prometheus-port` param, does it use 9090?

luislhl: That's a good point

pedroferreira1 (approved):

r4mmer (approved):

txstratum/cli.py
```diff
@@ -28,6 +28,7 @@ def create_parser() -> ArgumentParser:
     parser.add_argument('--tx-timeout', help='Tx mining timeout (seconds)', type=int, default=None)
     parser.add_argument('--fix-invalid-timestamp', action='store_true', help='Fix invalid timestamp to current time')
     parser.add_argument('--prometheus', help='Path to export metrics for Prometheus', type=str, default=None)
```
msbrogli, line 30: Maybe rename to `--prometheus-path`?

txstratum/prometheus.py
```diff
@@ -109,6 +88,62 @@ def _schedule_and_write_data(self) -> None:
             loop = asyncio.get_event_loop()
             loop.call_later(self.call_interval, self._schedule_and_write_data)
```
msbrogli, line 89: This sequence of calls is not reliable. Any unhandled exception in `update_metrics` will prevent the next schedule and thus prometheus will stop updating since then. I feel we should move to LoopingCall.

msbrogli (approved):

txstratum/cli.py
luislhl: I didn't do this to avoid adding a breaking change. Someone else could be using this arg out there and would break when upgrading.

txstratum/prometheus.py
luislhl: Done

Makefile
```diff
@@ -56,6 +56,8 @@ check: flake8 isort-check mypy
 
 # formatting:
 
+# TODO yapf formatting seems to be conflicting with flake8
+#      maybe we should tweak its configuration or replace it
```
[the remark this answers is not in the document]
luislhl: Created the issue: https://github.com/HathorNetwork/tx-mining-service/issues/46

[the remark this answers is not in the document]
luislhl: Skipping yapf in https://github.com/HathorNetwork/tx-mining-service/pull/44/commits/24756e38a91a23123b4ad3102b0643dd54fe1428

PR #45, opened 2021-12-13
feat: Allow logging in json

luislhl:
### Acceptance Criteria
- It should be possible to pass a flag that enables logs in JSON format

### DIFF
--- a/txstratum/cli.py
+++ b/txstratum/cli.py
@@ -16,13 +16,16 @@ from structlog import get_logger
 logger = get_logger()
 
 
+DEFAULT_LOGGING_CONFIG_FILE = "log.conf"
+
+
 def create_parser() -> ArgumentParser:
     """Create a parser for the cmdline arguments."""
     import configargparse  # type: ignore
     parser: ArgumentParser = configargparse.ArgumentParser(auto_env_var_prefix='hathor_')
     parser.add_argument('--stratum-port', help='Port of Stratum server', type=int, default=8000)
     parser.add_argument('--api-port', help='Port of TxMining API server', type=int, default=8080)
-    parser.add_argument('--log-config', help='Config file for logging', default='log.conf')
+    parser.add_argument('--log-config', help='Config file for logging', default=None)
     parser.add_argument('--json-logs', help='Enabled logging in json', default=False, action='store_true')
     parser.add_argument('--max-tx-weight', help='Maximum allowed tx weight to be mined', type=float, default=None)
     parser.add_argument('--max-timestamp-delta', help='Maximum allowed tx timestamp delta', type=int, default=None)
@@ -52,8 +55,14 @@ def execute(args: Namespace) -> None:
     from txstratum.utils import start_logging
 
     # Configure log.
+    log_config_file = args.log_config or DEFAULT_LOGGING_CONFIG_FILE
+
     start_logging()
     if args.json_logs:
+        if args.log_config:
+            logger.error('--json-logs and --log-config are mutually exclusive')
+            exit(1)
+
         logging.basicConfig(level=logging.INFO, format='%(message)s')
         from structlog.stdlib import LoggerFactory
         structlog.configure(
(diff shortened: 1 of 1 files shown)


andreabadesso (approved):

txstratum/cli.py
```diff
@@ -23,6 +23,7 @@ def create_parser() -> ArgumentParser:
     parser.add_argument('--stratum-port', help='Port of Stratum server', type=int, default=8000)
     parser.add_argument('--api-port', help='Port of TxMining API server', type=int, default=8080)
     parser.add_argument('--log-config', help='Config file for logging', default='log.conf')
+    parser.add_argument('--json-logs', help='Enabled logging in json', default=False, action='store_true')
```
jansegre, line 26: From what I understood of the implementation, this will make `--log-config` be completely ignored. I think we should either not allow both arguments to be set or at the very least log a warning/error when both are set (noting that `--log-config` will be ignored)

jansegre (changes requested):

txstratum/cli.py
```diff
@@ -52,7 +53,22 @@ def execute(args: Namespace) -> None:
 
     # Configure log.
     start_logging()
-    if os.path.exists(args.log_config):
+    if args.json_logs:
```
msbrogli, line 56: Why don't we use the log config file to enable JSON?

msbrogli (changes requested):

txstratum/cli.py
luislhl: Since all I needed in this case was `logging.basicConfig(level=logging.INFO, format='%(message)s')`, I thought it simple enough to stay in the code.

But I can check how to define another logger in the config file and choose which one to use.

txstratum/cli.py
luislhl: I'll improve this.

txstratum/cli.py
luislhl: Actually, now I'm not sure what your suggestion is exactly.

I don't think we could move the configuration I've added to a log config file, since most of what I've added is structlog's configuration. The only thing I would be able to move to a config file is this line: `logging.basicConfig(level=logging.INFO, format='%(message)s')`

Unless I'm missing something.

txstratum/cli.py
luislhl: > I think we should either not allow both arguments to be set 

Seems like the problem with this option is that `--log-config` has a default value. So it is always set.

I could check if it is set and different from default, and forbid this case. And log the warning/error every time `--json-logs` is set and `--log-config` is the default value.

What do you think?

luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/45/commits/f36fe89c1272e2568270bcf3d67aba6f0bd193fa

txstratum/cli.py
```diff
@@ -51,16 +55,37 @@ def execute(args: Namespace) -> None:
...
     # Configure log.
+    log_config_file = args.log_config or DEFAULT_LOGGING_CONFIG_FILE
+
     start_logging()
-    if os.path.exists(args.log_config):
-        logging.config.fileConfig(args.log_config)
+    if args.json_logs:
+        if args.log_config:
+            logger.error('--json-logs and --log-config are mutually exclusive')
+            exit(1)
```
jansegre, line 64: I guess this works, but I was thinking of using a `mutually_exclusive_group` like we use on hathor-core for example:

```python
        storage = parser.add_mutually_exclusive_group()
        storage.add_argument('--rocksdb-storage', action='store_true', help='Use RocksDB storage backend (default)')
        storage.add_argument('--memory-storage', action='store_true', help='Do not use any storage')
        storage.add_argument('--json-storage', action='store_true', help='Use legacy JSON storage (not recommended)')
```

txstratum/cli.py
jansegre: I though about this too, and I think I'm OK using a cli argument instead of using `log.conf` only because it's not that easy to figure out how to configure it. While I believe it is totally possible to use a `log.conf` it's maybe harder than it's worth.

jansegre (approved):

txstratum/cli.py
luislhl: Yeah, it's better. Done in https://github.com/HathorNetwork/tx-mining-service/pull/45/commits/23895a29822f46420f5f385ff46e4d2a52fac9e0

txstratum/cli.py
luislhl: > using a cli argument instead of using log.conf

I don't see how the log.conf alone would avoid the need for a cli argument.

My final objective is to give a choice to the user about logging in JSON or not.

So I need to know his choice somehow, because I need to add structlog's processors to configure my JSON logging. That's why I needed the cli argument.

As far as I could find in the docs and googling, it doesn't seem to be possible to add the structlog processors using only a log.conf. Is it possible?

The other way I see is to have another `log.conf`, like a `log_json.conf`. I could do this as well instead of the cli argument (which wouldn't avoid having to add the structlog processors, afaik)

jansegre: I think I understood now. Python's logging module may be completely configurable via `log.conf`, but `structlog.configure` does not load from `log.conf`. An alternative would be extending `log.conf` to support passing structlog configurations, but this seems complicated.

luislhl:
We have aligned with Marcelo that we can merge this.

PR #48, opened 2021-12-22
ci: push images to ECR

luislhl:
### Acceptance Criteria
- We should push images to ECR with tags of `dev-*`, `staging-*`, and `v*`. `staging-*` on commits to  `master`, `v*` when tags are created, and `dev-*` in any other case.
- We should only push images with tags `v*` to Dockerhub

### Motivation
We will be deploying a new instance of tx-mining-service that will use those images from ECR.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,8 +4,8 @@ on:
     tags:
     - v*
     # TODO: Remove
-    # branches:
-    # - ci/push-to-ecr
+    branches:
+    - ci/push-to-ecr
 
 jobs:
   # push-to-dockerhub:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,9 +3,7 @@ on:
   push:
     tags:
     - v*
-    # TODO: Remove
     branches:
-    - ci/push-to-ecr
     - master
 
 jobs:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,8 +3,35 @@ on:
   push:
     tags:
     - v*
+    # TODO: Remove
+    # branches:
+    # - ci/push-to-ecr
+
 jobs:
-  build:
+  # push-to-dockerhub:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #   - name: Set output
+  #     id: vars
+  #     run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+  #   - name: Set up QEMU
+  #     uses: docker/setup-qemu-action@27d0a4f181a40b142cce983c5393082c365d1480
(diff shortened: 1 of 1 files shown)


luislhl:
Maybe there is something wrong with tags. We got two untagged images and only one tagged: https://i.imgur.com/ngMtNul.png

luislhl:
> Maybe there is something wrong with tags. We got two untagged images and only one tagged: https://i.imgur.com/ngMtNul.png

I've commented about this in https://github.com/docker/build-push-action/issues/447#issuecomment-1005937581

.github/workflows/docker.yml
```diff
@@ -19,11 +47,17 @@ jobs:
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
+    - name: Login to AWS ECR
+      uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+      with:
+          registry: ${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.eu-central-1.amazonaws.com
```
jansegre, line 53: wouldn't it be better to use the whole URL as a parameter? something like `registry: ${{ secrets.AWS_ECR_URL }}`, that way it would be easier to change the ECR region

.github/workflows/docker.yml
```diff
@@ -3,13 +3,41 @@ on:
...
-      run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+    - name: Prepare tags
+      id: tags
+      shell: python
+      run: |
+        import datetime
+
+        timestamp = str(int(datetime.datetime.now().timestamp()))
+        base_ecr_tag = '${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.eu-central-1.amazonaws.com/tx-mining-service:'
+        base_dockerhub_tag = 'hathornetwork/tx-mining-service:'
```
jansegre, line 22: we could also take the chance to make this more explicit, which I think is a great bonus: `base_dockerhub_tag = 'docker.io/hathornetwork/tx-mining-service:'`

jansegre (approved):

.github/workflows/docker.yml
luislhl: Yeah, makes sense

.github/workflows/docker.yml
luislhl: Will do

.github/workflows/docker.yml
jansegre: Great!

.github/workflows/docker.yml
jansegre: Thanks!

r4mmer (approved):

PR #51, opened 2022-01-12
feat: avoid replacing tx jobs by themselves in the miners

luislhl:
### Acceptance Criteria

- We should rever replace a TxJob in a miner by the same TxJob

### Motivation
Currently, we try to refresh a miner job every `JOB_UPDATE_INTERVAL`, which is currently 2 seconds.

If the new job is a TxJob, we make the miner give up on its previous job when refreshing it, by sending it the `clean=True` parameter.

That means if a transaction took more than 2 seconds to be mined, it would kept being replaced by itself. In miners where the overhead of switching jobs is high, this would generate unnecessary overhead.

### Alternatives
Maybe just increasing `JOB_UPDATE_INTERVAL` would have a similar effect.


txstratum/protocol.py
```diff
@@ -345,12 +345,26 @@ def is_mining_block(self) -> bool:
...
+        """Refresh miner's job, updating timestamp.
+
+        This is run as a periodic task.
+        """
         assert self.current_job is not None
+
+        # If the miner is already mining a Tx, we do not interfere, otherwise we
+        # would be making it switch jobs unnecessarily.
+        if isinstance(self.current_job, MinerTxJob):
+            return
```
jansegre, line 357: What happens if the new job is also "mining a tx" but a different tx, what if the current_job was already found by another miner?

luislhl: > What happens if the new job is also "mining a tx" but a different tx

Yeah, we need to check this indeed. I'll change the implementation.

> what if the current_job was already found by another miner?

I think this would continue to work, since it is triggered in the manager method `stop_mining_tx`.
But we should discuss this after I commit the new implementation.

pedroferreira1 (approved):

pedroferreira1 (approved):

txstratum/protocol.py
jansegre: Great!

PR #53, opened 2022-01-12
feat: handle SIGTERM/SIGINT gracefully

luislhl:
### Acceptance Criteria
- When receiving a SIGTERM, the application should:
    - Wait for any current tx being mined, while refusing new ones
    - Ask all the connected miners to disconnect and reconnect, using the `client.reconnect` of the Stratum protocol
    - Refuse new connections from miners
- Github actions should run tests in all PRs (not just the ones with `dev` or `master` as base)
    
### Motivation
It's common in many systems, like container orchestrators, to send a SIGTERM signal to warn the application about an iminent SIGKILL, so it has time to finish whatever it needs, like closing connections or finishing jobs.

### Notes
We should also set the `--retry-pause` option in our miners to a lower value than default, which is 30 seconds, to make them reconnect faster after the connection is closed. Maybe 1s is a better value for this.

~I tested sending the `waittime` parameter described in https://en.bitcoin.it/wiki/Stratum_mining_protocol#client.reconnect, to make the miners wait a few seconds before reconnecting (which would give time for the new container to start), but it seems our miners don't really implement this.~ This was not needed after all, they reconnect instantly.

### TODO
- [ ] Make sure cpuminer obeys the `client.reconnect`message if we send no host/port
- [ ] Make sure ccminer obeys the `client.reconnect`message if we send no host/port
- [x] Make sure cgminer obeys the `client.reconnect`message if we send no host/port


.github/workflows/main.yml
```diff
@@ -7,9 +7,9 @@ on:
     tags:
     - v*
   pull_request:
-    branches:
-    - dev
-    - master
+    types:
+    - synchronize
+    - ready_for_review
```
jansegre, line 12: This is interesting. What do these mean? I still think it might make sense to run the CI on PRs that aren't ready for review, for example.

txstratum/cli.py
```diff
@@ -51,110 +53,154 @@ def create_parser() -> ArgumentParser:
...
+
+    def register_signal_handlers(self) -> None:
+        """Register signal handlers."""
+        import signal
+
+        logger.info('Registering signal handlers...')
+
+        sigterm = getattr(signal, 'SIGTERM', None)
+        if sigterm is not None:
+            self.loop.add_signal_handler(sigterm, self.sigterm_handler)
```
jansegre, line 86: Does this play well with SIGINT signal (or KeyboardInterrupt exception)? I think shutdown by Ctrl+C is useful too, so it might make sense to handle these too.

jansegre (approved):

txstratum/cli.py
luislhl: I can test adding a handler for SIGINT as well.

In Kubernetes it won't be possible to stop the process with Ctrl + C. The normal flow when a pod is deleted is that a SIGTERM is sent to the main process in the container, then a SIGKILL some seconds later. That's why I just added the SIGTERM handler.

But I guess we could have it handle SIGINT as well in cases where this is run in other kinds of infrastructure.

I will test this and make a new commit if it works well.

.github/workflows/main.yml
luislhl: I changed this because I noticed the workflow was not running for PRs that I had opened with a base that was not `dev` or `master`.

I got a explanation from their docs: 

`synchronize`: Triggered when a pull request's head branch is updated. For example, when the head branch is updated from the base branch, when new commits are pushed to the head branch, or when the base branch is changed. (https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request)

`ready_for_review`: A draft pull request was marked as ready for review. (https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types#ready_for_review)

But I'm not sure if `synchronize`  works for draft PRs.

luislhl: There are some people trying workarounds to avoid running jobs in Draft PRs, so I guess they actually run in Draft PRs as well with this config.

Example of discussion on workarounds: https://github.community/t/dont-run-actions-on-draft-pull-requests/16817/18

jansegre: So it looks like the workflow of PRs in draft might run because of `synchronize` then. This is what I prefer actually: to not prevent workflows from running just because they are drafts. But now I'm not sure what `ready_for_review` adds on top of `synchronize`, wouldn't `synchronize` be enough?

txstratum/cli.py
jansegre: I meant for development. Mostly because I think a similar approach will be useful for `hathor-core`, which already gracefully shutsdown on Ctrl+C but not on SIGTERM. (a Ctrl+C from the OS POV is just a SIGINT, but Python will "convert" it to a KeyboardInterrupt exception, I'm not too familiar with the specific mechanics of how this works though).

txstratum/protocol.py
```diff
@@ -184,6 +186,14 @@ def connection_lost(self, exc: Optional[Exception]) -> None:
         if self.refresh_job_task:
             asyncio.ensure_future(self.refresh_job_task.stop())
 
+    def ask_miner_to_reconnect(self) -> None:
+        """Ask the miner to reconnect. We just want to force it to disconnect with this."""
+        self.log.info('Asking miner to reconnect')
+        self.send_request('client.show_message', ['Will force reconnection'], None)
+        self.send_request('client.reconnect', [], None)
+
+        self.transport.close()
```
pedroferreira1, line 195: When `connection_made` is called? `self.transport` is not defined until that. 

Is it possible that this raise an exception depending on the moment it's called?

txstratum/cli.py
```diff
@@ -51,110 +53,154 @@ def create_parser() -> ArgumentParser:
...
-    loop.run_until_complete(server.wait_closed())
-    loop.run_until_complete(backend.stop())
-    loop.close()
+        logger.info('Stopping...')
+        for tx_filter in tx_filters:
+            loop.run_until_complete(tx_filter.close())
+        server.close()
+        loop.run_until_complete(server.wait_closed())
+        loop.run_until_complete(backend.stop())
+        loop.run_until_complete(manager.stop())
```
pedroferreira1, line 202: We were not stopping the manager before. Was it a bug?

txstratum/protocol.py
luislhl: It is called just after an instance of StratumProtocol is created, which happens in the manager:

```
    def __call__(self) -> StratumProtocol:
        """Return an instance of StratumProtocol for a new connection."""
        protocol = StratumProtocol(self)
```

But to be 100% honest with you, this is a mechanism I don't fully understand, because I never found where a Manager instance is called as a callable (which would trigger this __call__)

But by putting some logs I could confirm that `connection_made` is called immediately after the creation of a StratumProtocol instance.

So we will not close it before it is created.

txstratum/cli.py
luislhl: Yeah, I think so. But it didn't generate bad consequences, just a warning from asyncio that a task was still running (which is the task that the manager.stop method stops)

txstratum/cli.py
```diff
@@ -51,110 +53,154 @@ def create_parser() -> ArgumentParser:
...
+    """This is the main class of the service. It starts everything up."""
+
+    manager: TxMiningManager
+    loop: AbstractEventLoop
+
+    async def graceful_shutdown(self) -> None:
+        """Gracefully shutdown the service."""
+        logger.info('Gracefully shutting down...')
+
+        while len(self.manager.tx_queue) > 0:
```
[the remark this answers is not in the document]
luislhl: You are right, I moved something I had above this line and accidentally broke the logic. Will move it back.

txstratum/cli.py
luislhl: I've implemented handling of SIGINT in https://github.com/HathorNetwork/tx-mining-service/pull/53/commits/6f3d75d339936025f8e922430f629749b360f552

Python stopped converting it into a exception when I started handling it.

.github/workflows/main.yml
luislhl: Done

txstratum/cli.py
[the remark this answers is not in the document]
luislhl: Done

txstratum/cli.py
```diff
@@ -51,110 +57,172 @@ def create_parser() -> ArgumentParser:
...
+            logger.info('Waiting for pending txs to finish...', txs_left=len(self.manager.tx_queue))
+            await asyncio.sleep(2)
+
+        self.manager.shutdown()
+
+    def sigterm_handler(self) -> None:
+        """Handle SIGTERM signal."""
+        logger.info('SIGTERM received.')
+
+        self.loop.create_task(self.graceful_shutdown())
```
pedroferreira1, line 99: Any reason why sigterm calls `graceful_shutdown` and sigint `shutdown`? If that was a bug, maybe we could have a single method `handle_shutdown_signal`

luislhl: I guess you have a point.

I thought I had reasons to not run the full shutdown in Kubernetes, but after thinking about it again, I don't think there is any.

pedroferreira1 (approved):

.github/workflows/main.yml
jansegre: Great!

txstratum/cli.py
jansegre: Great, thanks!

jansegre (approved):
I just noticed that we "Wait for any current tx being mined, while refusing new ones". I don't think this is particularly bad, but it could make the time it takes from the SIGTERM being received to the process exiting vary a lot.

I don't think this is blocking or that it even goes against any official recommendation, but _usually_ processes will try to exit much faster when receiving a SIGTERM. I think this is mostly only relevant in the context of Kuberenetes, which is what I think originated this PR. I would imagine Kubernetes has a timeout strategy of sending a SIGKILL after a while of sending a SIGTERM if the process is still running, it could be that it sends more than one SIGTERM.

My only concern is if this plays well with whatever strategy Kubernetes uses.

Either way the current implementation is definitely progress in the sense that if there is any downside of taking too long to exist, it's probably rare. So if any of my concerns does not have a trivial answer, I'd rather have an issue to investigate it later.

luislhl:
> I just noticed that we "Wait for any current tx being mined, while refusing new ones". I don't think this is particularly bad, but it could make the time it takes from the SIGTERM being received to the process exiting vary a lot.
> 
> I don't think this is blocking or that it even goes against any official recommendation, but _usually_ processes will try to exit much faster when receiving a SIGTERM. I think this is mostly only relevant in the context of Kuberenetes, which is what I think originated this PR. I would imagine Kubernetes has a timeout strategy of sending a SIGKILL after a while of sending a SIGTERM if the process is still running, it could be that it sends more than one SIGTERM.
> 
> My only concern is if this plays well with whatever strategy Kubernetes uses.
> 
> Either way the current implementation is definitely progress in the sense that if there is any downside of taking too long to exist, it's probably rare. So if any of my concerns does not have a trivial answer, I'd rather have an issue to investigate it later.

In Kubernetes context we are safe with this implementation. The process of releasing a new version and replacing the old pod is this:

1. A new pod is created and starts receiving new requests.
2. The old pod is sent a SIGTERM, it will be able to respond pending requests and close connections
3. After 30s (by default) the old pod is sent a SIGKILL if it has not stopped yet

So we already have a new pod serving requests by the time the old pod receives the SIGTERM. (unless someone explicitly kills a pod, or it dies for any other reason, like being OOM killed, but these are edge cases)

There is no problem if the SIGTERM handling time varies from time to time.

But are you concerned only with Kubernetes, or with other environments as well?

jansegre:
> So we already have a new pod serving requests by the time the old pod receives the SIGTERM. (unless someone explicitly kills a pod, or it dies for any other reason, like being OOM killed, but these are edge cases)
> 
> There is no problem if the SIGTERM handling time varies from time to time.

Thanks!

> But are you concerned only with Kubernetes, or with other environments as well?

Yes. Because it's what we will use it's basically the only environment that matters right now.

PR #54, opened 2022-01-17
feat: more prometheus metrics

luislhl:
### Acceptance Criteria
We should have the following new prometheus metrics:
- Count of new tx jobs received
- Count of txs solved by miner address and miner type (cpuminers, ccminers and cgminers)
- Histogram of weight of all txs solved
- Histogram of weight of all txs timed out
- Histogram of the time it took to mine the txs, by miner type
- Histogram of the time the tx had to wait in the queue
- Count of completed jobs in each miner. This includes jobs that were not valid solutions. The objective of this is just to be able to detect miners that are connected, but are doing nothing.


txstratum/prometheus.py
```diff
@@ -16,21 +26,58 @@
...
+    ),
+    'txs_solved': Counter(
+        'txs_solved',
+        'Number of solved transactions',
+        labelnames=['miner_type', 'miner_address'],
+    ),
+    'txs_solved_weight': Histogram(
+        'txs_solved_weight',
+        'Txs solved histogram by tx weight',
+        buckets=(17, 18, 19, 20, 21, 22, 23, 25, float("inf"))
```
luislhl, line 50: We should add a label here to identify the type of miner that solved them

jansegre (approved):
LGTM. I only have a question that if generates any problem, could be dealt with on a new PR.

txstratum/protocol.py
```diff
@@ -237,22 +254,25 @@ def method_submit(self, params: Any, msgid: JSONRPCId) -> None:
...
+            # If the solution is not valid, get a new job.
             self.manager.update_miner_job(self, clean=True)
 
     def method_configure(self, params: Any, msgid: JSONRPCId) -> None:
         """Handle stratum-extensions configuration from JSONRPC.
 
         See: https://github.com/slushpool/stratumprotocol/blob/master/stratum-extensions.mediawiki
         """
-        self.log.debug('handle configure', msgid=msgid, params=params)
+        self.log.debug('handle conaboutfigure', msgid=msgid, params=params)
```
pedroferreira1, line 275: Typo?

txstratum/prometheus.py
```diff
@@ -16,21 +26,59 @@
...
+    ),
+    'txs_solved': Counter(
+        'txs_solved',
+        'Number of solved transactions',
+        labelnames=['miner_type', 'miner_address'],
+    ),
+    'txs_solved_weight': Histogram(
+        'txs_solved_weight',
+        'Txs solved histogram by tx weight',
+        buckets=(17, 18, 19, 20, 21, 22, 23, 25, float("inf")),
```
pedroferreira1, line 50: Why no 24?

txstratum/prometheus.py
```diff
@@ -16,21 +26,59 @@
...
+        labelnames=['miner_type']
+    ),
+    'txs_timeout_weight': Histogram(
+        'txs_timeout_weight',
+        'Txs timeouts histogram by tx weight',
+        buckets=(17, 18, 19, 20, 21, 22, 23, 25, float("inf"))
```
pedroferreira1, line 56: Why no 24?

txstratum/prometheus.py
```diff
@@ -43,7 +91,6 @@ def collect_metrics(manager: 'TxMiningManager') -> MetricData:
     return MetricData(
         miners_count=len(manager.miners),
         total_hashrate_ghs=manager.get_total_hashrate_ghs(),
-        txs_solved=manager.txs_solved,
```
pedroferreira1, line 46: I was thinking about keeping the global txs solved count. Is it bad?

pedroferreira1 (approved):

txstratum/protocol.py
luislhl: Yes

txstratum/prometheus.py
luislhl: Once I made an analysis of distribution of weights in the mainnet, and according to it we didn't have many txs in this weight zone.

So I thought we could just have a larger bucket without losing too much information.

The more buckets we have the more data we will generate in Prometheus, so I tried just creating what is necessary.

But I guess we could add the 24 if you think it's necessary, one more won't kill us.

histogram

txstratum/prometheus.py
luislhl: Same reason as in https://github.com/HathorNetwork/tx-mining-service/pull/54#discussion_r787137761

txstratum/prometheus.py
luislhl: We will have the total count from the Histogram metric, this one would be redundant.

txstratum/pubsub.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def emit(self, event: TxMiningEvents, obj: Any) -> None:
+        """Emit an event.
+
+        :param event: event to emit
+        :param obj: object that will be passed to subscriber coroutines
+        """
+        for callable in self._subscribers[event]:
+            coroutine = callable(obj)
+            self.loop.create_task(coroutine)
```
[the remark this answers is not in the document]
jansegre: Looks great! Thanks!

txstratum/prometheus.py
pedroferreira1: Ah, ok. I thought the new one was just aggregating by miner, so the total count could be different, given that two different miners could solve the same tx.

But if we have the two different metrics, fine by me

txstratum/prometheus.py
pedroferreira1: I think it's fine to skip 24

txstratum/prometheus.py
pedroferreira1: As we discussed privatly, this is fine

PR #55, opened 2022-01-20
fix: get miner version from cgminers

luislhl:
### Acceptance Criteria
We should be able to get the miner version from cgminers.

Previously we couldn't, because they only accept integers as ids for messages, and we were using a string as id. This is what causes this behavior there: https://github.com/HathorNetwork/cgminer/blob/master/util.c#L2468

To achieve this, his PR implements a better way to manage the messages that are waiting return from the miners.

### Notes
I didn't test this with `ccminer` miners. But we shouldn't have problems since their implementation to handle the `get_version` message is almost the same as the `cpuminer` one. This can be seen in https://github.com/HathorNetwork/ccminer/blob/master/util.cpp#L1683 and https://github.com/HathorNetwork/cpuminer/blob/master/util.c#L1408

### DIFF
--- a/txstratum/protocol.py
+++ b/txstratum/protocol.py
@@ -3,7 +3,6 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 import asyncio
-import random
 import uuid
 from math import log2
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, NamedTuple, Optional, cast
@@ -143,7 +142,7 @@ class StratumProtocol(JSONRPCProtocol):
         else:
             self.log.error('Cant handle result: {}'.format(result), miner_id=self.miner_id, msgid=msgid)
 
-    def get_next_message_id(self):
+    def get_next_message_id(self) -> int:
         """Return the next message id."""
         msgid = self.next_message_id
         self.next_message_id += 1
--- a/txstratum/protocol.py
+++ b/txstratum/protocol.py
@@ -3,6 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 import asyncio
+import random
 import uuid
 from math import log2
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, NamedTuple, Optional, cast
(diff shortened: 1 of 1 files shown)


txstratum/protocol.py
```diff
@@ -106,18 +119,45 @@ def __init__(self, manager: 'TxMiningManager'):
...
+
+    def send_and_track_request(self, method: str, params: Any) -> None:
+        """Send a request to the client and track it."""
+        assert len(self.messages_in_transit) < 1000000
+
+        msgid = random.randint(1, 1000000)
+
+        # We need to make sure msg ids are unique
+        while msgid in self.messages_in_transit:
+            msgid = random.randint(1, 1000000)
```
jansegre, line 153: I don't think you need a random it, a sequential id works fine, like how we do on the merged-mining coordinator: https://github.com/HathorNetwork/hathor-core/blob/master/hathor/merged_mining/coordinator.py#L342-L347

But either, maybe just making this into a "next_msgid" method would be good enough.

jansegre (approved):

txstratum/protocol.py
```diff
@@ -156,6 +196,25 @@ def start_periodic_tasks(self) -> None:
...
+        now = txstratum.time.time()
+
+        should_delete = []
+
+        for message in self.messages_in_transit.values():
+            if now > message.timeout:
+                should_delete.append(message.id)
+
+        for msgid in should_delete:
+            message = self.messages_in_transit.pop(msgid)
```
pedroferreira1, line 215: With this timeout we might never have the client version if it times out?

Should we have a retry mechanism for it? 

We are using the version "only" for saving metrics in a better way, right?

pedroferreira1 (approved):

txstratum/protocol.py
luislhl: > With this timeout we might never have the client version if it times out?
The timeout happens after 30 seconds, which is a lot of time.

I created it just to make sure we wouldn't keep them in memory.
We can even increase the time.

If the miner really didn't respond in this time, then we surely had some other problem that would happen even if we waited more time.

> Should we have a retry mechanism for it?
> 
> We are using the version "only" for saving metrics in a better way, right?

Yeah, only for metrics. And if we can't get the version for some reason, we will just get an `unknown` in Prometheus tags.

I think we are safe enough in this case that a retry isn't that important.
But maybe in the future, if we send more critical messages using this mechanism.

txstratum/protocol.py
luislhl: Done

jansegre (approved):

txstratum/protocol.py
pedroferreira1: Yeah, I agree

PR #62, opened 2022-02-23
ci: replace yapf with black

luislhl:
### Acceptance Criteria
- We should use `black` instead of `yapf` to auto-format the files
- We should remove deprecated options from `isort` (see https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0.html)
- We should run black in all files so it formats them

### Motivation
yapf was conflicting with flake8, some changes made by it were caught by flake8 as errors.


jansegre (approved):
LGTM, looking forward to doing this to more Python projects. Should we also add a CI step to check the formatting?

Makefile
```diff
@@ -45,7 +45,7 @@ mypy-tests: $(py_tests)
 
 .PHONY: flake8
 flake8: $(py_sources) $(py_tests)
-	flake8 $^
+	flake8 --max-line-length=121 $^
```
r4mmer, line 48: Should this be on setup.cfg?

pyproject.toml
```diff
@@ -14,7 +14,7 @@ homepage = "https://hathor.network/"
 repository = "https://github.com/HathorNetwork/tx-mining-service/"
 
 [tool.poetry.dependencies]
-python = ">=3.6,<4.0"
+python = ">=3.7,<4.0"
```
r4mmer, line 17: Since we only build for python 3.9 (Fixed on the Dockerfile), why not make this `>=3.9,<4.0`?

pyproject.toml
```diff
@@ -35,7 +35,16 @@ numpy = "^1.19.4"
...
+black = "^22.1.0"
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.black]
+line-length = 120
+
+[tool.isort]
```
r4mmer, line 47: I think we should use pyproject or setup.cfg for this configuration. isort has configuration on both files and this can cause unexpected behaviors. Maybe configure all (including black) on setup.cfg or bring the configurations there to pyproject

r4mmer (changes requested):
Unrelated to this PR, but I saw that the `clean` rule on Makefile uses `clean-protos` which does not exist, maybe we should remove it as well?

pyproject.toml
luislhl: I'm ok with this, but this would require people to have Python 3.9 in their local environments to run the project.

I'm not sure this would not pose a challenge to someone.

What do you think?

pyproject.toml
luislhl: You are right, I didn't notice we had setup.cfg already

pyproject.toml
r4mmer: If the PR validation builds and tests on 3.9 I don't think it's an issue.

r4mmer (approved):

PR #64, opened 2022-03-03
feat: rate limits

luislhl:
### Acceptance Criteria
- All APIs will have a rate limit of 1 request per second per IP
- It should be possible to choose between Redis or Memory as storage for the rate limits with the `--rate-limits-storage` option
- It should possible to configure the Redis url with the `--redis-url` option
- The rate limits will be disabled by default, and will be enabled only if `--rate-limits-storage` option is passed.

### Notes
This is part of https://github.com/HathorNetwork/internal-issues/issues/79

### DIFF
--- a/tests/test_rate_limiter.py
+++ b/tests/test_rate_limiter.py
@@ -1,13 +1,6 @@
-from asyncore import loop
-
-import asynctest
+import asynctest  # type: ignore[import]
 from aiohttp import web
-from aiohttp.test_utils import (
-    AioHTTPTestCase,
-    TestClient,
-    TestServer,
-    unittest_run_loop,
-)
+from aiohttp.test_utils import TestClient, TestServer
 
 from txstratum.rate_limiter import MemoryLimiter, get_default_keyfunc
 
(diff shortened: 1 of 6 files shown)


tests/test_rate_limiter.py
```diff
@@ -0,0 +1,48 @@
...
+        self.assertTrue(data["success"])
+
+        # Second call should be rate limited
+        resp = await self.client.request("GET", "/route")
+        assert resp.status == 429
+        data = await resp.json()
+        self.assertEqual(data["error"], "Rate limit exceeded: 1 per 3 second")
+
+        # Clear the rate limiter
+        self.fake_app.clear()
```
jansegre, line 42: Shouldn't we also test a "wait" in addition to a "clear".

jansegre (approved):

tests/test_rate_limiter.py
luislhl: You mean to simulate the time passing and then making a new request?

That's something I tried for a while instead of this `clear` solution.

But after spending some time and not succeeding, I decided it was enough to simulate what happens in the lib when the time passes, which is something similar to what this `clear` does:
- clear: https://github.com/alisaifee/limits/blob/master/limits/storage/memory.py#L84
- expiration with time: https://github.com/alisaifee/limits/blob/master/limits/storage/memory.py#L37

But I can also try to spend some more time trying to simulate the time passing in the test, to force the library to expire the events.

luislhl:
Closing this because a new design was approved to handle this with Nginx

PR #70, opened 2022-03-30
docs: run in production

luislhl:
Design: https://github.com/HathorNetwork/internal-issues/issues/96

### Acceptance Criteria
- Guide about running a tx-mining-service in production
- Guide about running the miners
- Guide about the api limits

### Notes
Maybe our `cgminer` repo is too old and will not support newer ASIC devices, because of lack of drivers.

Should we think about updating it?

### DIFF
--- a/docs/apidocs.yml
+++ b/docs/apidocs.yml
@@ -4,7 +4,7 @@ servers:
 info:
   title: tx-mining-service API
   description: This service is used to resolving transactions before propagating them into the network
-  version: 0.2.0
+  version: 0.6.2
 paths:
   /health-check:
     get:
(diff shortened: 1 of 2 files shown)


docs/running_in_production.md
```diff
@@ -0,0 +1,256 @@
...
+You can find more info about running fullnodes in https://hathor.gitbook.io/hathor/guides/full-node
+
+### Miners
+You will need miners to mine the transactions.
+
+We have a dedicated guide on how to run them in [/docs/running_miners.md](/docs/running_miners.md).
+
+### Reverse Proxy / API Gateway
+
+We suggested that you use a reverse proxy or API Gateway in front of the service to handle the API calls.
```
luislhl, line 23: suggested -> suggest

Issue #69, opened 2022-03-28
Return success on repeated jobs (instead of return an error)


msbrogli:
When the push tx fails and one tries to send the same transaction, this new attempt will fail in the tx-mining-service step because of the tx-mining-service will return error "job-already-exists".

This issue proposes to change the return to a success with the job-id when it receive a job that already exists. It should return something very similar to a new job.

We need a simple design to discuss each field before writing an acceptance criteria for this issue.

PR #71, opened 2022-03-31
feat: return success on repeated job submission

r4mmer:
# Acceptance criteria

- [ ] Implement the behavior described on #69 
- [ ] Clients should behave the same way when a successful job submission and a repeated job submission is issued

### DIFF
--- a/txstratum/api.py
+++ b/txstratum/api.py
@@ -157,7 +157,8 @@ class App:
             self.manager.add_job(job)
         except JobAlreadyExists:
             self.log.debug("job-already-exists", data=data)
-            return web.json_response({"error": "job-already-exists"}, status=400)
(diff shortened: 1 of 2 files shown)


txstratum/api.py
```diff
@@ -157,7 +157,8 @@ async def submit_job(self, request: web.Request) -> web.Response:
             self.manager.add_job(job)
         except JobAlreadyExists:
             self.log.debug("job-already-exists", data=data)
-            return web.json_response({"error": "job-already-exists"}, status=400)
```
msbrogli, line 160: It seems the current set of tests does not cover this part. I feel we should add tests to it.

msbrogli (changes requested):

msbrogli (approved):

PR #76, opened 2022-04-13
fix: handle hathor-core errors when submitting blocks

luislhl:
closes: https://github.com/HathorNetwork/tx-mining-service/issues/77

depends on: https://github.com/HathorNetwork/python-hathorlib/pull/10

### Acceptance Criteria
- Tx Mining Service should not stop pushing blocks in case one of the pushes gets an error from the fullnode
- It should try again for 3 times to push a transaction if it gets error from the fullnode
- It should just ignore errors when pushing blocks

### Notes
Linters are failing currently because this code depends on the new hathorlib. This PR is blocked by the lib PR

### TODO
- [x] Use the new lib when it's released

### DIFF
--- a/tests/test_manager.py
+++ b/tests/test_manager.py
@@ -14,6 +14,7 @@ from unittest.mock import ANY, MagicMock, Mock
 import asynctest  # type: ignore
 import pytest
 from hathorlib.client import BlockTemplate, HathorClient
+from hathorlib.exceptions import PushTxFailed
 
 import txstratum.time
 from txstratum.exceptions import JobAlreadyExists
@@ -780,6 +781,9 @@ class ManagerTestCase(unittest.TestCase):
         conn.send_error = MagicMock(return_value=None)
         conn.send_result = MagicMock(return_value=None)
         conn.method_submit(params=params, msgid=None)
+
+        self._run_all_pending_events()
+
         conn.send_error.assert_not_called()
         conn.send_result.assert_called_once_with(None, "ok")
 
@@ -841,7 +845,7 @@ class ManagerTestCase(unittest.TestCase):
 
             async def push_tx_or_block(self, raw: bytes) -> bool:
                 self.pushes_attempted += 1
-                raise RuntimeError("Error pushing block")
+                raise PushTxFailed("Error pushing block")
 
         self.manager.backend = BrokenHathorClient(server_url="")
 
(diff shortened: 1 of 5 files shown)


tests/test_manager.py
```diff
@@ -409,6 +409,9 @@ def test_two_miners_same_submission_1(self):
         conn1.send_result = MagicMock(return_value=None)
         self.manager.backend.push_tx_or_block = MagicMock(return_value=asyncio.Future())
         conn1.method_submit(params=params, msgid=None)
+
+        self._run_all_pending_events()
```
luislhl, line 413: `self.manager.backend.push_tx_or_block` is now called from inside a async method in the manager. So I needed to add this here (and below) to make sure this async method would have been already called before the assertions.

tests/test_manager.py
```diff
@@ -148,6 +148,9 @@ async def _fn():
         future = asyncio.ensure_future(_fn())
         self.loop.run_until_complete(future)
 
+    def _get_ready_miner(self, address: Optional[str] = None) -> StratumProtocol:
```
luislhl, line 151: Moves this just because it was kind of hidden in the middle of the tests cases.

tests/test_manager.py
```diff
@@ -822,6 +831,56 @@ def test_miner_block_submission_after_receiving_tx(self):
...
+        conn.send_result.reset_mock()
+        block_job.submitted_at = None
+        conn.method_submit(params=params, msgid=None)
+        conn.send_error.assert_not_called()
+        conn.send_result.assert_called_once_with(None, "ok")
+
+        self._run_all_pending_events()
+
+        # Assert that the push was really tried again
+        self.assertEqual(self.manager.backend.pushes_attempted, 2)
```
luislhl, line 882: Before the bug was fixed, this assertion would fail, because we wouldn't try to push blocks anymore. Making sure a new push is tried even after the error from the fullnode asserts that the bug was fixed.

txstratum/manager.py
```diff
@@ -40,6 +40,9 @@ class TxMiningManager:
     DEFAULT_BLOCK_TEMPLATE_UPDATE_INTERVAL = 3.0  # seconds
     TX_CLEAN_UP_INTERVAL = 300.0  # seconds
 
+    MAX_RETRY_FOR_FAILED_PUSH_TX = 3
+    RETRY_WAIT_FOR_FAILED_PUSH_TX = 3  # seconds
```
jansegre, line 44: Maybe this is too slow? This will affect the time a job takes to be processed right?

jansegre (approved):
It looks good to me, although it also comes with the feature of retrying a submit if it fails.

txstratum/manager.py
luislhl: Yeah, if we consider that "processing" the job includes pushing the transaction (if they have propagate=true), then yes, we would increase the time.

For wallets that use the tx-mining-service, I think they wouldn't be affected by this, because they propagate the transactions themselves, not on tx-mining-service (maybe I should confirm this).

But the reason why I didn't set a lower time is to give the fullnode some time to recover before trying again, if the error was just due to temporary load.

What do you think?

jansegre: If the default is for wallets to push the transactions themselves, I think a retry mechanism is really not that relevant, if it fails it fails and we move on, blocks found by the tx-mining-service aren't urgently important, and in the few cases they are actually needed (like the testnet), they will be mined and found again, just at a later timestamp.

But for the cases that opt to let the tx-mining-service push the transactions, I'd say it is unexpected for a job to succeed and the tx not be pushed, so maybe a better behavior would be to wait for the push of the transaction to mark the job as complete.

jansegre (approved):
Looks great to me!

tests/test_manager.py
```diff
@@ -847,6 +952,16 @@ def tearDown(self):
     def _get_ready_miner(self, address: Optional[str] = None) -> StratumProtocol:
         return _get_ready_miner(self.manager, address)
 
+    async def _run_all_pending_events(self):
```
msbrogli, line 955: That's clever! I feel we can use it in hathor-core as well.

tests/test_manager.py
```diff
@@ -958,6 +1073,7 @@ async def test_job_mining_time(self):
         conn.send_error = MagicMock(return_value=None)
         conn.send_result = MagicMock(return_value=None)
         conn.method_submit(params=params, msgid=None)
+        self._run_all_pending_events()
```
luislhl, line 1076: I think I missed an `await` here :thinking:

PR #81, opened 2022-06-07
feat: add min_version parameter for all three wallets and block requests from old ones

pedroferreira1:
### Acceptance Criteria

- We must be able to block from old wallet versions. This is done getting the wallet version from the user agent and comparing with the minimum allowed version set in the command line argument.
- This should be possible for wallet desktop, mobile and headless.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -123,11 +123,9 @@ def get_timestamp(tx_bytes: bytes) -> int:
     return tx.timestamp
 
 
-class AppTestCase(AioHTTPTestCase):
-    async def get_application(self):
-        self.manager = TxMiningManager(backend=None, pubsub=MagicMock(), address=None)
-        self.myapp = App(self.manager)
-        return self.myapp.app
+class BaseAppTestCase(AioHTTPTestCase):
+    __test__ = False
+    version_check: bool
 
     @unittest_run_loop
     async def test_health_check(self):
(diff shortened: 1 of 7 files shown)


txstratum/middleware.py
```diff
@@ -0,0 +1,103 @@
...
+            # Then check if version is the minimum allowed
+            # Example of a wallet desktop user agent
+            # Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) HathorWallet/0.22.1
+            # Chrome/73.0.3683.121 Electron/5.0.13 Safari/537.36 HathorWallet/0.22.1
+            wallet_desktop_regex = r"HathorWallet/(\d+\.\d+.\d+)"
+            if is_version_invalid(
+                wallet_desktop_regex, user_agent, min_wallet_desktop_version
+            ):
+                return web.json_response(
+                    {"error": "wallet-version-invalid"}, status=400
```
msbrogli, line 46: Maybe `"wallet-version-too-old"`?

txstratum/utils.py
```diff
@@ -404,3 +404,23 @@ def _exception_handler(loop: "AbstractEventLoop", context: Dict[str, Any]) -> No
...
+    """Compare two versions in the format X.Y.Z.
+
+    Validates if version is greater than or equal to base_version
+    """
+    version_split = version.split(".")
+    base_version_split = base_version.split(".")
+
+    assert len(version_split) == len(base_version_split)
+
+    for i in range(len(version_split)):
```
msbrogli, line 419: I'm afraid you are comparing strings instead of numbers. Maybe just create a `Tuple[int]` and `return version_split >= base_version_split`.

msbrogli (changes requested):

txstratum/middleware.py
pedroferreira1: I agree it's better. Done here https://github.com/HathorNetwork/tx-mining-service/pull/81/commits/61fb9ca35fdba7a7f03c4c3511c2545688fa1e38

txstratum/utils.py
pedroferreira1: You are right. The tuple comparison is interesting, I had forgotten about that, thanks. 

Done here https://github.com/HathorNetwork/tx-mining-service/pull/81/commits/58f622b0d3ac716ca06f0ad4ffe0d23dac6a5d89

txstratum/middleware.py
```diff
@@ -0,0 +1,107 @@
...
+
+    return version_check
+
+
+def is_version_invalid(regex: str, user_agent: str, min_version: str) -> bool:
+    """Check if version in user agent is invalid comparing with min version.
+
+    Regex is used to get the version from user agent
+    """
+    search = re.search(regex, user_agent)
```
luislhl, line 102: It is possible that compiling the regex beforehand (outside this function) would be more efficient. By using `re.search` every time, we could be compiling then every time as well.

https://docs.python.org/3/library/re.html#re.compile

Although for this case that has just a few regex'es I'm not really sure it makes any difference, because apparently the `re` module caches some of the compiled regex'es.

But I would say we should compile them explicitly even if there is no gain right now, to make it explicit that we are trying to optimize, and because we don't know how many more regex'es we could add in the future

txstratum/middleware.py
```diff
@@ -0,0 +1,107 @@
...
+            # Then check if version is the minimum allowed
+            # Example of a wallet desktop user agent
+            # Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) HathorWallet/0.22.1
+            # Chrome/73.0.3683.121 Electron/5.0.13 Safari/537.36 HathorWallet/0.22.1
+            wallet_desktop_regex = r"HathorWallet/(\d+\.\d+.\d+)"
+            if is_version_invalid(
+                wallet_desktop_regex, user_agent, min_wallet_desktop_version
+            ):
+                return web.json_response(
+                    {"error": VERSION_CHECK_ERROR_MESSAGE}, status=400
```
luislhl, line 50: Should we add the detected version and the minimum expected version in the response?

Maybe it is useful in some cases, especially in the wallet-headless, that could be logging the received response.

txstratum/cli.py
```diff
@@ -109,6 +109,27 @@ def create_parser() -> ArgumentParser:
...
+
+    parser.add_argument(
+        "--min-wallet-mobile-version",
+        help="Minimum version for the wallet mobile to use this tx mining",
+        type=str,
+        default=None,
+    )
+
+    parser.add_argument(
+        "--min-wallet-headless-version",
```
luislhl, line 127: We will need a PR in our infra as well to set these arguments.

txstratum/utils.py
```diff
@@ -404,3 +404,23 @@ def _exception_handler(loop: "AbstractEventLoop", context: Dict[str, Any]) -> No
...
+    # into tuples of integers
+
+    try:
+        version_int_tuple = tuple(map(int, version.split(".")))
+        base_version_int_tuple = tuple(map(int, base_version.split(".")))
+    except ValueError:
+        # version strings are not in the expected format
+        return False
+
+    assert len(version_int_tuple) == len(base_version_int_tuple)
```
msbrogli, line 424: I feel we should have an if and raise an exception (instead of using an assert).

tests/test_api.py
```diff
@@ -486,3 +485,157 @@ async def test_submit_job_fail_script_data26_non_standard(self):
...
+            "/submit-job",
+            json={"tx": tx_hex},
+            headers=header_mobile_custom_wrong,
+        )
+        await resp.json()
+        self.assertEqual(200, resp.status)
+
+        header_headless_lower = {"User-Agent": "Hathor Wallet Headless / 0.14.87"}
+        header_headless_equal = {"User-Agent": "Hathor Wallet Headless / 0.14.88"}
+        header_headless_higher = {"User-Agent": "Hathor Wallet Headless / 0.14.89"}
```
msbrogli, line 591: Have you added tests for invalid versions? For example: Hathor Wallet Headless / 0.xx.88 or Hathor Wallet Headless / 0.14.88beta?

txstratum/middleware.py
pedroferreira1: I can do it but I feel wouldn't help much to be honest, each `re.search` I do is for a different regex, so I can't reuse them. I think this would be helpful if we were running many searches for a single regex, which is not the case.

Anyway, I can do it no problem. Do you feel it's worth it compiling now even with different regex to prevent loss of performance in the future?

tests/test_api.py
pedroferreira1: Done https://github.com/HathorNetwork/tx-mining-service/pull/81/commits/2c6c6d6a74a8af506c02d9839f0ecacb4a463c2f

txstratum/middleware.py
pedroferreira1: I was trying to keep the error string as an "error code", similar to what we have in other responses in the tx mining service but I agree it's useful for logging. 

I've added in another dict field the versions, what do you think? https://github.com/HathorNetwork/tx-mining-service/pull/81/commits/60dd9040f8286ac11ff30a72b038a08678d46bed

txstratum/middleware.py
```diff
@@ -0,0 +1,107 @@
...
+            response = await handler(request)
+            return response
+
+        if min_wallet_desktop_version:
+            # Search user agent for wallet desktop string and get version
+            # Then check if version is the minimum allowed
+            # Example of a wallet desktop user agent
+            # Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) HathorWallet/0.22.1
+            # Chrome/73.0.3683.121 Electron/5.0.13 Safari/537.36 HathorWallet/0.22.1
+            wallet_desktop_regex = r"HathorWallet/(\d+\.\d+.\d+)"
```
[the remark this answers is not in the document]
luislhl: :heavy_check_mark:

txstratum/middleware.py
luislhl: Great

luislhl (approved):

tests/test_api.py
msbrogli: Thanks!

msbrogli (approved):

PR #84, opened 2022-06-15
Fix/rust build

luislhl:


jansegre:
I suggest switching the `python` base image from `-alpine` to `-slim`, which is Debian based. Mainly because it's easier and faster to just use the official wheels provided by cryptography, which use manylinux, which, if I'm not mistaken, works on Debian but not on Alpine.

I had a similar issue on `hathor-core`, however since we recently switched to a Debian base image it is now only affecting pypy+arm builds, because there aren't official wheels for those.

The other alternative I think would be to set-up our own runners, which we can make more beefy and also we can have arm runners that will run the arm tasks (although to make that work with the Docker build we'd have to make a few changes). But even just running our own runners seems like a big undertaking already.

luislhl:
> I suggest switching the `python` base image from `-alpine` to `-slim`, which is Debian based. Mainly because it's easier and faster to just use the official wheels provided by cryptography, which use manylinux, which, if I'm not mistaken, works on Debian but not on Alpine.

Yeah, it makes sense to try to use the official wheels. But while learning about this, I saw that apparently they provide wheels for `musllinux` as well, not only `manylinux`: https://cryptography.io/en/latest/installation/#static-wheels

`musllinux` wheels should support Alpine.

Apparently our build was not using them just because Poetry was not prepared yet to use `musllinux` wheels: https://github.com/python-poetry/poetry/issues/5225

After upgrading to Poetry `1.2.0b2`, the build was able to use the wheels as expected.

So I think this is enough to solve it. What do you think?

Dockerfile
```diff
@@ -6,12 +6,12 @@ WORKDIR /code
 RUN apk add --no-cache gcc musl-dev libffi-dev openssl-dev rust cargo
 
 RUN pip --no-input --no-cache-dir install --upgrade pip wheel
-RUN pip --no-input --no-cache-dir install 'poetry==1.1.6'
+RUN pip --no-input --no-cache-dir install 'poetry==1.2.0b2'
```
jansegre, line 9: I think it might be better to not pin a specific version of poetry but instead use the latest version, like we do with pip/wheel. At least that's how it is on `hathor-core`, mostly because we had issues with things stopping to work for not using the latest poetry, at least that happens more frequently than it breaking for using the latest version (which I believe has never happened).

.github/workflows/docker.yml
```diff
@@ -56,7 +57,7 @@ jobs:
     - name: Build and push
       uses: docker/build-push-action@a66e35b9cbcf4ad0ea91ffcaf7bbad63ad9e0229
       with:
-        push: true
+        push: false
```
jansegre, line 60: Is this supposed to have changed? Or is it just for testing?

If it's of any use on `hathor-core` there is an example of building the image, testing if the image works, and only then pushing the image.

jansegre (approved):

Dockerfile
luislhl: I will try `'poetry>=1.2.0b2'`, like you suggested on Slack

.github/workflows/docker.yml
```diff
@@ -5,6 +5,7 @@ on:
     - v*
     branches:
     - master
+    - fix/rust-build
```
pedroferreira1, line 8: Just as a TODO to remove this before the merge

pedroferreira1 (approved):

PR #98, opened 2023-06-01
chore: update dependencies

pedroferreira1:
## Acceptance Criteria

- Upgrade python hathorlib to v0.4.0. This version adds support to the serialization/deserialization of nano contract objects. With that, the transaction mining service can mine nano contract transactions.
- Upgrade other dependencies.
- Fix mypy error with method that was not returning the type set.


tests/test_manager.py
```diff
@@ -97,7 +97,7 @@ async def stop(self):
     async def get_block_template(self, address: Optional[str] = None) -> BlockTemplate:
         return self._block_templates[self._current_index]
 
-    async def get_tx_parents(self) -> List[bytes]:
+    async def get_tx_parents(self):
```
jansegre, line 100: Is this not used? Seems like a method that should return something, what about `raise NotImplementedError` instead of changing the method signature?

jansegre (approved):

tests/test_manager.py
pedroferreira1: Makes sense! This is a method from a class we are extending from the python hathorlib. 

I've rolled back my changes and added the `raise NotImplementedError`, much better. Thanks

msbrogli (approved):

PR #104, opened 2023-09-28
feat: health check

luislhl:
### Acceptance Criteria

We should have a new `/health` endpoint, that will implement a more complex health check which includes:

- Checking that there is at least 1 miner connected
- Checking that the miners have at least 1 job submitted in the last hour (submitting a job is different than mining a transaction, there should be submitted jobs even if no transactions are being mined)
- Checking that none of the jobs in the last 5 minutes failed
- Checking that the jobs aren't taking more than 10 seconds to be mined
- Checking the health of the Fullnode
    - For now we will only make sure its `/v1a/version` endpoint is responding. In future PRs we will use the new yet-to-be-implemented `/v1a/health` endpoint of hathor-core.

### DIFF
--- a/txstratum/healthcheck.py
+++ b/txstratum/healthcheck.py
@@ -1,0 +1,209 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
+
+if TYPE_CHECKING:
+    from txstratum.manager import TxMiningManager
+
+
+class ComponentType(str, Enum):
+    """
+    Enum used to store the component types that can be used in the HealthCheckComponentStatus class.
+    """
+    DATASTORE = 'datastore'
+    INTERNAL_COMPONENT = 'internal_component'
+
+
+class HealthCheckStatus(str, Enum):
+    """
+    Enum used to store the component status that can be used in the HealthCheckComponentStatus class.
+    """
+    PASS = 'pass'
+    WARN = 'warn'
+    FAIL = 'fail'
+
+
+@dataclass
+class ComponentHealthCheck:
+    """
+    This class is used to store the result of a health check in a specific component.
+    """
+    component_name: str
+    component_type: ComponentType
+    status: HealthCheckStatus
+    output: str
+    time: Optional[str] = None
+    component_id: Optional[str] = None
+    observed_value: Optional[str] = None
+    observed_unit: Optional[str] = None
+
+    def update(self, new_values: Dict[str, Any]) -> None:
+        """
+        Update the object with the new values passed as kwargs.
+        Also updates the time field with the current time with the format YYYY-MM-DDTHH:mm:ssZ
+        """
+        self.time = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
+
+        for key, value in new_values.items():
+            setattr(self, key, value)
+
+    def to_json(self) -> dict:
+        """
+        Return a dict representation of the object. All field names are converted to camel case.
+        """
+        json = {
+            'componentType': self.component_type.value,
+            'status': self.status.value,
+            'output': self.output
+        }
+
+        if self.component_id:
+            json['componentId'] = self.component_id
+
+        if self.observed_value:
+            assert self.observed_unit is not None, 'observed_unit must be set if observed_value is set'
+
+            json['observedValue'] = self.observed_value
+            json['observedUnit'] = self.observed_unit
+
+        return json
+
+
+@dataclass
+class HealthCheckResult:
+    status: HealthCheckStatus
+    description: str
+    checks: Dict[str, List[ComponentHealthCheck]]
+
+    def get_http_status_code(self) -> int:
+        """
+        Return the HTTP status code for the status.
+        """
+        if self.status == HealthCheckStatus.PASS:
+            return 200
+        elif self.status == HealthCheckStatus.WARN:
+            return 503
+        elif self.status == HealthCheckStatus.FAIL:
+            return 503
+        else:
+            raise ValueError('Invalid status')
+
+    def to_json(self) -> dict:
+        """
+        Return a dict representation of the object. All field names are converted to camel case.
+        """
+        return {
+            'status': self.status.value,
+            'description': self.description,
+            'checks': {k: [c.to_json() for c in v] for k, v in self.checks.items()}
+        }
+
+
+class HealthCheckInterface(ABC):
+    """
+    This is an interface to be used by other classes implementing health checks for components.
+    """
+    @abstractmethod
+    def get_health_check(self) -> ComponentHealthCheck:
+        """
+        Return the health check status for the component.
+        """
+        raise NotImplementedError()
+
+
+class HealthCheck(HealthCheckInterface):
+    """This is the main class that will use the other classes to check the health of the components.
+    It will aggregate the responses into a final object to be returned following our standards.
+    """
+    def __init__(self, manager: "TxMiningManager") -> None:
+        self.health_check_components: List[HealthCheckInterface] = [
+            ManagerHealthCheck(manager)
+        ]
+
+    def get_health_check(self) -> HealthCheckResult:
+        components_health_checks = [c.get_health_check() for c in self.health_check_components]
+        components_status = {c.status for c in components_health_checks}
+
+        overall_status = HealthCheckStatus.PASS
+        if HealthCheckStatus.FAIL in components_status:
+            overall_status = HealthCheckStatus.FAIL
+        elif HealthCheckStatus.WARN in components_status:
+            overall_status = HealthCheckStatus.WARN
+
+        return HealthCheckResult(
+            status=overall_status,
+            description='health of txstratum service',
+            checks={c.component_name: [c] for c in components_health_checks}
+        )
+
+
+# class FullnodeHealthCheck(HealthCheckInterface):
+#     """
+#     This class will check the health of the fullnode by sending a request to its /v1a/health
+#     """
+
+
+class ManagerHealthCheck(HealthCheckInterface):
+    """
+    This class receives a manager instance as parameter and implements a health check method for it that will:
+    - Check that the manager has at least 1 miner
+    - Check that all 'tx_jobs' in the manager in the last 5 minutes have a status of done
+      and have a total_time lesser than 10 seconds.
+
+    If at least one of the 'tx_jobs' has a 'total_time' of more than 10 seconds, the health check
+    will be returned as 'warn'.
+
+    If some of the 'tx_jobs' has status different than 'done', the health check will be returned
+    as 'fail'.
+
+    If there are no miners, the health check will be returned as 'fail'
+    """
+    def __init__(self, manager: "TxMiningManager") -> None:
+        self.manager = manager
+        self.last_manager_status = ComponentHealthCheck(
+            component_name='manager',
+            component_type=ComponentType.INTERNAL_COMPONENT,
+            status=HealthCheckStatus.PASS,
+            output='everything is ok'
+        )
+
+    def get_health_check(self) -> ComponentHealthCheck:
+        """
+        Return the manager health check status.
+        """
+        if not self.manager.has_any_miner():
+            self.last_manager_status.update({
+                'status': HealthCheckStatus.FAIL,
+                'output': 'no miners connected'
+            })
+
+            return self.last_manager_status
+
+        if not self.manager.tx_jobs:
+            # We just return the last status in case there are no jobs in the last 5 minutes
+            return self.last_manager_status
+
+        for job in self.manager.tx_jobs.values():
+            if job.is_failed():
+                self.last_manager_status.update({
+                    'status': HealthCheckStatus.FAIL,
+                    'output': 'some tx_jobs in the last 5 minutes have failed'
+                })
+
+                return self.last_manager_status
+            if job.total_time > 10:
+                self.last_manager_status.update({
+                    'status': HealthCheckStatus.WARN,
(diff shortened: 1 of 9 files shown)


txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+        self.health_check_components: List[ComponentHealthCheckInterface] = [
+            MiningHealthCheck(manager),
+            FullnodeHealthCheck(backend),
+        ]
+
+    async def get_health_check(self) -> HealthCheckResult:
+        """Return the health check status for the tx-mining-service."""
+        components_health_checks = [
+            await c.get_health_check() for c in self.health_check_components
+        ]
```
glevco, line 132: You could use [`asyncio.gather()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) here to run all futures concurrently. Probably won't change much in practice, but it's nice

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+        components_status = {c.status for c in components_health_checks}
+
+        overall_status = HealthCheckStatus.PASS
+        if HealthCheckStatus.FAIL in components_status:
+            overall_status = HealthCheckStatus.FAIL
+        elif HealthCheckStatus.WARN in components_status:
+            overall_status = HealthCheckStatus.WARN
```
glevco, line 139: If `HealthCheckStatus` was an `IntEnum` instead of an `StrEnum`, you could assign enum values such that this would simply be `overall_status = min(components_status)`. And then that could probably be an utility static method in `HealthCheckStatus`.

Could be good if we ever end up adding new status values (we wouldn't need to remember to add it here)

txstratum/healthcheck.py
glevco, line 1: What do you think about moving some classes to their own files? Maybe not all of them, but at least separating model classes from "checkers"

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+        try:
+            # TODO: We need to get the health information from the fullnode, but it's not implemented yet
+            await self.backend.version()
+
+            return health_check
+        except Exception:
+            health_check.status = HealthCheckStatus.FAIL
+            health_check.output = "couldn't connect to fullnode"
+
+            return health_check
```
glevco, line 176: Could we somehow carry the failure information to the health check result? For example, the status code and response content could be helpful in debugging.

glevco, line 176: Also, I'm not sure how the `HathorClient` class works, does it perform any check on the returned value? I mean, I know that some full node APIs can respond with 200 OK even if it fails, and the contents could be something like `{"success": false}`.

Maybe we should check some attribute that we know is returned from `/v1a/version`, just to be sure. For example, checking that we receive a `version` attribute (only while `/v1a/health` is not implemented).

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+        if not self.manager.has_any_miner():
+            return ComponentHealthCheck(
+                component_name="manager",
+                component_type=ComponentType.INTERNAL,
+                status=HealthCheckStatus.FAIL,
+                output="no miners connected",
+            )
+
+        # Check that all the miners submitted in the last 1 hour. If not, return failed.
+        if not self.manager.has_any_submitted_job_in_period(period=3600):
```
glevco, line 224: Move 3600 to constant?

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+    """This class will check the health of the fullnode by sending a request to its /v1a/health."""
+
+    def __init__(self, backend: "HathorClient") -> None:
+        """Init the class with the fullnode backend."""
+        self.backend = backend
+
+    async def get_health_check(self) -> ComponentHealthCheck:
+        """Return the fullnode health check status."""
+        health_check = ComponentHealthCheck(
+            component_name="fullnode",
```
glevco, line 158: Maybe move component names to a constant, or even better an enum?

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+            return self.last_manager_status
+
+        for job in self.manager.tx_jobs.values():
+            if job.is_failed():
+                self.last_manager_status.update(
+                    {
+                        "status": HealthCheckStatus.FAIL,
+                        "output": "some tx_jobs in the last 5 minutes have failed",
+                    }
+                )
```
glevco, line 243: Maybe changing `update()` to receive `**new_values` would provide a nicer interface (you wouldn't have to create a dict here)

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+
+                return self.last_manager_status
```
glevco, line 245: If I'm not mistaken, if we have a failed job and then stay, say, 10 minutes without any new jobs, we would respond with the `"some tx_jobs in the last 5 minutes have failed"` message even though there were no jobs in the last 5 minutes, which could be misleading during debugging.

Not sure if just changing this message is enough. It would probably better to create a new status for the "no jobs in the last 5 minutes" case, adding a specific output message.

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+            if job.is_done():
+                assert (
+                    job.total_time is not None
+                ), "total_time must be set if job is done"
+
+                if job.total_time > 10:
```
glevco, line 251: Move 10 to a constant?

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+                component_type=ComponentType.INTERNAL,
+                status=HealthCheckStatus.FAIL,
+                output="no miners submitted a job in the last 1 hour",
+            )
+
+        if not self.manager.tx_jobs:
+            # We just return the last saved status in case there are no jobs in the last 5 minutes
+            return self.last_manager_status
+
+        for job in self.manager.tx_jobs.values():
```
glevco, line 236: If there's a failed job _after_ a long running job in the array, we would emit `WARN` even though we should emit `FAIL`.

We could just fix this logic or, as a suggestion, we could change it a bit to send an aggregation of values. For example, "we had X failed jobs and Y long-running jobs in the last 5 minutes". That second part could be a future improvement, though.

glevco (commented):
Overall great PR, very well documented. I'm leaving some comments but most of them are optional.

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+    observed_value: Optional[str] = None
+    observed_unit: Optional[str] = None
+
+    def update(self, new_values: Dict[str, Any]) -> None:
+        """
+        Update the object with the new values passed as kwargs.
+
+        Also updates the time field with the current time with the format YYYY-MM-DDTHH:mm:ssZ
+        """
+        self.time = datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
```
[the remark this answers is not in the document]
luislhl: That's a good question, I had to check this.

Apparently this custom time module exists just to facilitate controlling the time during tests.

For the health check logic it doesn't make much different which one we use.

But since they are using it, I'll refactor the health check to use the same, so we keep things consistent.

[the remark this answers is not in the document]
luislhl: That's a good question, I had to check this.

Apparently they use this custom time module just to be able to easily control time during tests.

For the health check logic this is not needed, and would just add some additional complexity if we try to use their module, since they only wrap the `time.time()` function. I would need the `.gmtime()` and `.strftime()` as well, which I would be able to get only from their `txstratum._time`, which seems ugly.

So I would say we can safely keep the `datetime.utcnow()` in this case as the simplest solution.

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+
+        for key, value in new_values.items():
+            setattr(self, key, value)
```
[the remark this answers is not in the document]
luislhl: Do you see any advantage?

txstratum/healthcheck.py
luislhl: We will implement a health endpoint for hathor-core as well, which will provide us with detailed information in case the fullnode is unhealthy. This code will be refactor to use this new endpoint when it's ready.

But for now I'll make a change to include the Exception information in the output, which could help in debugging.

luislhl: You can see the `HathorClient` in https://github.com/HathorNetwork/python-hathorlib/blob/main/hathorlib/client.py

The class just assumes the "happy path" and tries to extract the `version` attribute from the response and parse it.

So it kind of already checks the attribute, and if it's not there some kind of Exception will be raised and we will catch it.

I was considering this to be good enough while we don't implement the `/v1a/health` on hathor-core, which should happen soon.

txstratum/healthcheck.py
luislhl: I've fixed this in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/379470f4c215148a0b13fec8699af290a719a3ed

txstratum/healthcheck.py
luislhl: Fixed in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/4958a03b9290bf85f05d19a1faebcc5fddc004b3

txstratum/healthcheck.py
luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/bf397c3dc89f5b5471df0188f97fed4c064acf2d

txstratum/healthcheck.py
luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/acc340e690cfd4bf508fa18b6b25a2a98385b5aa

txstratum/healthcheck.py
luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/acc340e690cfd4bf508fa18b6b25a2a98385b5aa

txstratum/healthcheck.py
luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/acc340e690cfd4bf508fa18b6b25a2a98385b5aa

txstratum/healthcheck.py
luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/cab88bb366cda5e2af5019f3037268bc9a452364

txstratum/healthcheck.py
luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/d7880ffb00149c1584a54c4effae123ecd2e6d85

txstratum/healthcheck.py
```diff
@@ -0,0 +1,265 @@
...
+    def get_http_status_code(self) -> int:
+        """Return the HTTP status code for the status."""
+        if self.status == HealthCheckStatus.PASS:
+            return 200
+        elif self.status == HealthCheckStatus.WARN:
+            return 503
+        elif self.status == HealthCheckStatus.FAIL:
+            return 503
+        else:
+            raise ValueError("Invalid status")
```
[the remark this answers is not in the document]
luislhl: Done in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/1c25e43785b8a4db7c9c3e00d67f72b654cada17

txstratum/healthcheck.py
luislhl: After some changes I think this is better now, although I didn't change to `IntEnum`. Can you check it?

txstratum/healthcheck.py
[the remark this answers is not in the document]
glevco: Well, I was hoping `replace()` could help with type safety, but I tested it and it doesn't. In this case, it would only simplify the code a bit, as now it could simply be `replace(self, **new_values)` instead of the loop. But then you would have to return the new instance instead of updating the current one in place, which I do think it's better (i always prefer immutability), but would require a small refactor in you code and I don't see it as necessary.

glevco (approved):

txstratum/healthcheck/healthcheck.py
```diff
@@ -0,0 +1,184 @@
...
+    """This class will check the health of the fullnode by sending a request to its /v1a/health."""
+
+    def __init__(self, backend: "HathorClient") -> None:
+        """Init the class with the fullnode backend."""
+        self.backend = backend
+
+    async def get_health_check(self) -> ComponentHealthCheck:
+        """Return the fullnode health check status."""
+        health_check = ComponentHealthCheck(
+            component_name="fullnode",
```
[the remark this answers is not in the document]
luislhl: fixed in https://github.com/HathorNetwork/tx-mining-service/pull/104/commits/e6cc2be5d73908cf2e73abbf31fa7711aaea0b79

tuliomir (approved):

txstratum/healthcheck/healthcheck.py
```diff
@@ -0,0 +1,186 @@
...
+    def __init__(self, backend: "HathorClient") -> None:
+        """Init the class with the fullnode backend."""
+        self.backend = backend
+
+    async def get_health_check(self) -> ComponentHealthCheck:
+        """Return the fullnode health check status."""
+        health_check = ComponentHealthCheck(
+            component_name=self.COMPONENT_NAME,
+            component_type=ComponentType.FULLNODE,
+            # TODO: Ideally we should not use private fields. We'll fix this when fixing line 170
```
luislhl, line 66:
```suggestion
            # TODO: Ideally we should not use private fields. We'll fix this when fixing line 74 below about getting the health information from the fullnode.
```

PR #106, opened 2023-11-27
docs: add healthcheck

luislhl:
### Acceptance Criteria

We should add the new `/health` endpoint in the API docs.

The endpoint was implemented in https://github.com/HathorNetwork/tx-mining-service/pull/104


glevco (approved):

tuliomir (approved):
I would just suggest adding a link to #104 on the description to emphasize the relation with that feature.

PR #114, opened 2024-10-22
feat: implement task to estimate RTT value for each miner

pedroferreira1:
### Description

One possible improvement for our hashrate estimator is to consider the RTT value in the time to mine the job. Before we change the code to consider this rtt value, the first step is to calculate it and check if it influences the final hashrate estimation.

### Acceptance Criteria

- Add RTT method estimator.
- Store RTT value for each miner in prometheus.

### DIFF
--- a/tests/fixtures/expected_prometheus_metrics.prom
+++ b/tests/fixtures/expected_prometheus_metrics.prom
@@ -21,8 +21,8 @@ tx_queue 2.0
 block_template_error 1.0
 # HELP miner_rtt Miner RTT
 # TYPE miner_rtt gauge
-miner_rtt{miner_address="abc123"} 0.0
-miner_rtt{miner_address="def456"} 0.0
+miner_rtt{miner_address="abc123",miner_id="id1"} 0.0
+miner_rtt{miner_address="def456",miner_id="id2"} 0.0
 # HELP txs_jobs_received_total Number of transactions received by the API
 # TYPE txs_jobs_received_total counter
 txs_jobs_received_total 0.0
@@ -31,10 +31,10 @@ txs_jobs_received_total 0.0
 txs_jobs_received_created 1.642511672616624e+09
 # HELP txs_solved_total Number of solved transactions
 # TYPE txs_solved_total counter
-txs_solved_total{miner_address="abc123",miner_type="cpuminer"} 1.0
+txs_solved_total{miner_address="abc123",miner_id="id1",miner_type="cpuminer"} 1.0
 # HELP txs_solved_created Number of solved transactions
 # TYPE txs_solved_created gauge
-txs_solved_created{miner_address="abc123",miner_type="cpuminer"} 1.6425116730388892e+09
+txs_solved_created{miner_address="abc123",miner_id="id2",miner_type="cpuminer"} 1.6425116730388892e+09
 # HELP txs_solved_weight Txs solved histogram by tx weight
 # TYPE txs_solved_weight histogram
 txs_solved_weight_bucket{le="17.0",miner_type="cpuminer"} 0.0
--- a/tests/fixtures/expected_prometheus_metrics.prom
+++ b/tests/fixtures/expected_prometheus_metrics.prom
@@ -19,6 +19,10 @@ tx_queue 2.0
 # HELP block_template_error Number of errors updating block template
 # TYPE block_template_error gauge
 block_template_error 1.0
+# HELP miner_rtt Miner RTT
+# TYPE miner_rtt gauge
+miner_rtt{miner_address="abc123"} 0.0
+miner_rtt{miner_address="def456"} 0.0
 # HELP txs_jobs_received_total Number of transactions received by the API
 # TYPE txs_jobs_received_total counter
 txs_jobs_received_total 0.0
--- a/txstratum/protocol.py
+++ b/txstratum/protocol.py
@@ -51,6 +51,7 @@ class MessageInTransit(NamedTuple):
     id: int  # Some miners, like cgminer, only support integer ids
     method: str
     timeout: float
+    created_at: float
 
 
 class StratumProtocol(JSONRPCProtocol):
@@ -67,6 +68,9 @@ class StratumProtocol(JSONRPCProtocol):
     JOB_UPDATE_INTERVAL = 2  # in seconds, to update the timestamp of the job
     MESSAGES_TIMEOUT_INTERVAL = (
         5  # in seconds, the interval to clear timed out messages
+    )
+    RTT_ESTIMATOR_INTERVAL = (
+        5  # in seconds, to estimate the RTT from application to miner
     )
 
     MIN_WEIGHT = 20.0  # minimum "difficulty" to assign to jobs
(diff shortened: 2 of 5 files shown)


tests/fixtures/expected_prometheus_metrics.prom
```diff
@@ -19,6 +19,10 @@ tx_queue 2.0
 # HELP block_template_error Number of errors updating block template
 # TYPE block_template_error gauge
 block_template_error 1.0
+# HELP miner_rtt Miner RTT
+# TYPE miner_rtt gauge
+miner_rtt{miner_address="abc123"} 0.0
```
msbrogli, line 24: Address as in wallet address? Or uid? I guess there might be multiple connections using the same wallet address which might make things confusing.

txstratum/protocol.py
```diff
@@ -151,8 +158,17 @@ def handle_result(self, result: Any, msgid: JSONRPCId) -> None:
...
+            if self.miner_version is None:
+                # Logs only the first time, otherwise this log would appear a lot because
+                # we use this method to estimate the RTT
+                self.log.info(
+                    "Miner version: {}".format(result), miner_id=self.miner_id
+                )
             self.miner_version = result
+            now = txstratum.time.time()
+            current_rtt = now - message.created_at
+            alpha = 0.7
```
msbrogli, line 170: Maybe move to a constant at the class level.

txstratum/protocol.py
```diff
@@ -169,11 +185,10 @@ def get_next_message_id(self) -> int:
     def send_and_track_request(self, method: str, params: Any) -> None:
         """Send a request to the client and track it."""
         msgid = self.get_next_message_id()
+        now = txstratum.time.time()
 
         self.messages_in_transit[msgid] = MessageInTransit(
-            id=msgid,
-            method=method,
-            timeout=txstratum.time.time() + self.MESSAGE_TIMEOUT,
+            id=msgid, method=method, timeout=now + self.MESSAGE_TIMEOUT, created_at=now
```
msbrogli, line 191: Why not one parameter per line?

txstratum/protocol.py
```diff
@@ -234,6 +249,17 @@ def start_periodic_tasks(self) -> None:
...
+        if self.rtt_estimator_task is None:
+            # Start rtt estimator periodic task.
+            self.rtt_estimator_task = Periodic(
+                self.rtt_estimator, self.RTT_ESTIMATOR_INTERVAL
+            )
+            asyncio.ensure_future(self.rtt_estimator_task.start())
+
+    async def rtt_estimator(self) -> None:
+        """Period task to request the miner version to calculate the rtt."""
+        self.send_and_track_request("client.get_version", None)
```
msbrogli, line 261: Should we check whether the last request was fulfilled? I guess we shouldn't have two requests waiting for reply.

msbrogli (approved):

txstratum/protocol.py
pedroferreira1: The linter did it but I can change again

tests/fixtures/expected_prometheus_metrics.prom
luislhl: I guess we can add the `miner_id` as another label to all metrics that currently use only `miner_address`.

Adding labels won't disrupt our charts in Grafana.

txstratum/protocol.py
luislhl: The interval of this RTT estimator is only 5 seconds, while the timeout of the message sent to the miners is 30 seconds.

This opens a room for this possibility to happen.

luislhl (approved):

tests/fixtures/expected_prometheus_metrics.prom
pedroferreira1: Done to all metrics that use `miner_address`

https://github.com/HathorNetwork/tx-mining-service/pull/114/commits/094b75aa15affb9d878eef615e32162f3749cf52

txstratum/protocol.py
pedroferreira1: Done https://github.com/HathorNetwork/tx-mining-service/pull/114/commits/413e9e2f80260e45b8f23f3b876d6f975599ce69

txstratum/protocol.py
pedroferreira1: Done https://github.com/HathorNetwork/tx-mining-service/pull/114/commits/413e9e2f80260e45b8f23f3b876d6f975599ce69

txstratum/protocol.py
pedroferreira1: I changed it back but the linter complained again

txstratum/protocol.py
```diff
@@ -151,8 +160,18 @@ def handle_result(self, result: Any, msgid: JSONRPCId) -> None:
...
+                # we use this method to estimate the RTT
+                self.log.info(
+                    "Miner version: {}".format(result), miner_id=self.miner_id
+                )
             self.miner_version = result
+            now = txstratum.time.time()
+            current_rtt = now - message.created_at
+            alpha = self.RTT_CONFIDENCE_CONSTANT
+            self.rtt = (1 - alpha) * self.rtt + alpha * current_rtt
+            self.rtt_message_in_transit = False
```
luislhl, line 174: I think there is some chance that we could never get a reply from the miner, right?

In this case, we would never set this to False, and consequently we would stopped sending new messages to it.

Should we use a date for this message instead of a boolean, to avoid this possibility?

We would set it to None here, and set it to the current date when sending a new one, and when checking this variable, we would check if it's not None and the date in it is not too old.

Does this make sense?

luislhl (approved):

txstratum/protocol.py
pedroferreira1: Yes, it makes sense. Added a timeout of 60s https://github.com/HathorNetwork/tx-mining-service/pull/114/commits/9f8f4be1493e5cbe0892b4c63dc9a50bf23e0602

txstratum/protocol.py
```diff
@@ -68,6 +69,13 @@ class StratumProtocol(JSONRPCProtocol):
     MESSAGES_TIMEOUT_INTERVAL = (
         5  # in seconds, the interval to clear timed out messages
     )
+    RTT_ESTIMATOR_INTERVAL = (
+        5  # in seconds, to estimate the RTT from application to miner
+    )
+    RTT_ESTIMATOR_MESSAGE_TIMEOUT = (
+        60  # in seconds, the maximum time to wait for an rtt message to be received
+    )
+    RTT_CONFIDENCE_CONSTANT = 0.7
```
msbrogli, line 78: I'd call it `RTT_ALPHA_SLIDING_WINDOW`

msbrogli (approved):

txstratum/protocol.py
msbrogli: Great!

txstratum/protocol.py
msbrogli: Thanks!

tests/fixtures/expected_prometheus_metrics.prom
msbrogli: Thanks!

txstratum/protocol.py
pedroferreira1: Done https://github.com/HathorNetwork/tx-mining-service/pull/114/commits/a8ffaedd2d48fe5a75a0a22ae1322d2ebbff82ca

PR #115, opened 2024-10-22
feat: improve data collection of metrics

pedroferreira1:
### Acceptance Criteria

- Reduce initial weight for hashrate estimator because cpuminers were taking too long to mine their first job.
- Increase the quantity of buckets for `txs_solved_weight` and `txs_timeout_weight` metrics, so we can analyze better the distribution.
- Add log to collect data for the miner's hashrate that is solving each transaction and block.
- Fix log

### DIFF
--- a/txstratum/prometheus.py
+++ b/txstratum/prometheus.py
@@ -47,13 +47,13 @@ METRICS_PUBSUB = {
     "txs_solved_weight": Histogram(
         "txs_solved_weight",
         "Txs solved histogram by tx weight",
-        buckets=(17, 18, 19, 20, 21, 22, 23, 25, 26, 27, 28, 29, 30, float("inf")),
+        buckets=(17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, float("inf")),
         labelnames=["miner_type"],
     ),
     "txs_timeout_weight": Histogram(
         "txs_timeout_weight",
         "Txs timeouts histogram by tx weight",
-        buckets=(17, 18, 19, 20, 21, 22, 23, 25, 26, 27, 28, 29, 30, float("inf")),
+        buckets=(17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, float("inf")),
     ),
     "txs_mining_time": Histogram(
         "txs_mining_time",
--- a/txstratum/prometheus.py
+++ b/txstratum/prometheus.py
@@ -47,13 +47,13 @@ METRICS_PUBSUB = {
     "txs_solved_weight": Histogram(
         "txs_solved_weight",
         "Txs solved histogram by tx weight",
-        buckets=(17, 18, 19, 20, 21, 22, 23, 25, float("inf")),
+        buckets=(17, 18, 19, 20, 21, 22, 23, 25, 26, 27, 28, 29, 30, float("inf")),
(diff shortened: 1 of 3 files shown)


txstratum/prometheus.py
```diff
@@ -47,13 +47,13 @@
     "txs_solved_weight": Histogram(
         "txs_solved_weight",
         "Txs solved histogram by tx weight",
-        buckets=(17, 18, 19, 20, 21, 22, 23, 25, float("inf")),
+        buckets=(17, 18, 19, 20, 21, 22, 23, 25, 26, 27, 28, 29, 30, float("inf")),
```
luislhl, line 50: Should we add 24 as well? Since apparently you wanted a step of 1

txstratum/protocol.py
```diff
@@ -372,8 +372,22 @@ def method_submit(self, params: Any, msgid: JSONRPCId) -> None:
...
+                    "Block solved: ",
+                    miner_id=self.miner_id,
+                    miner_type=self.miner_type,
+                    hashrate=self.hashrate_ghs,
+                    weight=job.get_weight(),
+                )
                 self.blocks_found += 1
             else:
+                self.log.info(
+                    "Transaction solved: ",
```
luislhl, line 385: This can be logged twice for the same transaction, if the miner sends two solutions for it before it receives a new job.

We've seen this happening with the `ccminer` in the wild.

If you're going to use these logs for any additional aggregation and analysis later, would this influence in your analysis?

txstratum/prometheus.py
pedroferreira1: You are correct, I didn't notice this gap. Done https://github.com/HathorNetwork/tx-mining-service/pull/115/commits/6dd95496da371543d95e57c529eac07dd3c05d6a

txstratum/protocol.py
pedroferreira1: This is fine. The information we want to collect is the miner that is able to solve this weight (and its hashrate in that moment)

luislhl (approved):

txstratum/protocol.py
```diff
@@ -372,8 +372,22 @@ def method_submit(self, params: Any, msgid: JSONRPCId) -> None:
             # to solve the block.
             self.manager.submit_solution(self, job, obj.get_struct_nonce())
             if obj.is_block:
+                self.log.info(
+                    "Block solved: ",
+                    miner_id=self.miner_id,
```
msbrogli, line 377: I guess you could add the block hash so you can filter out duplicated log entries.

txstratum/protocol.py
msbrogli: I guess you could add the tx hash so you can filter out duplicated log entries.

msbrogli (approved):

txstratum/protocol.py
pedroferreira1: Done https://github.com/HathorNetwork/tx-mining-service/pull/115/commits/6a03c3952974bf811598bdc1cbfcdc149ed9584d

txstratum/protocol.py
pedroferreira1: Done https://github.com/HathorNetwork/tx-mining-service/pull/115/commits/6a03c3952974bf811598bdc1cbfcdc149ed9584d

PR #123, opened 2024-12-20
fix: add OPTIONS /submit-job endpoint for CORS

glevco:
### Motivation

I had issues with CORS when trying to run a wallet-desktop with a local tx-mining-server running in docker. This PR fixes it. For more info, see [this comment](https://github.com/HathorNetwork/hathor-wallet/pull/710#discussion_r1894441789).

### Acceptance Criteria

- Add `OPTIONS /submit-job` endpoint returning an empty success response.

### DIFF
--- a/txstratum/api.py
+++ b/txstratum/api.py
@@ -83,6 +83,7 @@ class App:
         self.app.router.add_get("/health", self.health)
         self.app.router.add_get("/mining-status", self.mining_status)
         self.app.router.add_get("/job-status", self.job_status)
+        self.app.router.add_options("/submit-job", self.submit_job_options)
         self.app.router.add_post("/submit-job", self.submit_job)
         self.app.router.add_post("/cancel-job", self.cancel_job)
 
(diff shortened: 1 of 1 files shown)


tuliomir (approved):
In theory, the `on_prepare` method should be enough for the `OPTIONS` to work, but I can attest that the error on the Desktop Wallet requires this change.

txstratum/api.py
```diff
@@ -83,6 +83,7 @@ def __init__(
         self.app.router.add_get("/health", self.health)
         self.app.router.add_get("/mining-status", self.mining_status)
         self.app.router.add_get("/job-status", self.job_status)
+        self.app.router.add_options("/submit-job", self.submit_job_options)
```
pedroferreira1, line 86: Should we also add the options to the `cancel-job` POST?

In this case, I would call the options method `get_options_response` or something like that. What do you think?

glevco: I don't see any reason in doing this, as it's not used the (current) QA. I think we should either add it to only the endpoints we use, or to all of them (so when the QA evolves, it just works).

glevco: Done in https://github.com/HathorNetwork/tx-mining-service/pull/123/commits/eb0d9ea561addb3bfc2b021f169aa2273e26bf1e

PR #127, opened 2025-01-23
fix: poetry build on docker

glevco:
### Motivation

The docker build was failing because poetry v2 was released, requiring a change in our config. We now have to explicitly tell poetry that our project is not a package ([docs](https://python-poetry.org/docs/basic-usage/#operating-modes)).

### Acceptance Criteria

- Explicitly set `package-mode = false` in the `pyproject.toml`.
- Fix a minor warning in the `Dockerfile`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-alpine as build
+FROM python:3.9-alpine AS build
 
 WORKDIR /code
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@
 [tool.poetry]
 name = "tx-mining-service"
 version = "0.14.0"
+package-mode = false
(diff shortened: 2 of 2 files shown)


tuliomir (approved):

tuliomir:
Tangentially related to this PR, should we also upgrade the Dockerfile in the line below, to explicitly require v2 or greater?

https://github.com/HathorNetwork/tx-mining-service/blob/591548c1a3a9076be16e17a4c5613183bcc4b3b6/Dockerfile#L9

glevco:
@tuliomir 

> Tangentially related to this PR, should we also upgrade the Dockerfile in the line below, to explicitly require v2 or greater?

Poetry v2 was released just a few days ago, so I think we could wait a bit more before forcing it as the minimum version.

pedroferreira1 (approved):

